PR #7, opened 2015-02-06
Removed test1 and test2 sirens

waylon531:
I removed the names for the 2 asian sirens, they should default to siren now


rvanherel:
Awesome stuff!

PR #8, opened 2015-02-07
Fixed window bug

### LINKED ISSUE #3: Look mode: windowed walls and GLTerrain mention.
The ground type is mentioned when examining a wall with a window, but not when examining a windowless wall:

screen shot 2014-06-27 at 22 32 59

waylon531:
I couldn't find the root cause of the bug, but I was able to add a check to see if the tile had a window or not. This may cause problems if there are walls that need to be able to see under, but I don't think that is a likely case. Fixes #3

### DIFF
--- a/Main/Source/lsquare.cpp
+++ b/Main/Source/lsquare.cpp
@@ -710,7 +710,7 @@ void lsquare::UpdateMemorizedDescription(truth Cheat)
     {
       MemorizedDescription.Empty();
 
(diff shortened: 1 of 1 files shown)


emlai:
@waylon531 Just a thought: Could `OLTerrain->IsWall()` be used here instead?

waylon531:
I didn't notice that when I was looking at olterrain, but apparently it doesn't work.

waylon531:
Derp, when I was testing IsWall() I hadn't flipped the boolean, making only walls show the floor tile. It works when I actually use it correctly.

PR #21, opened 2015-03-03
Breakable Whips

Clawzfire:
Makes whips breakable, since they were somewhat overpowered when they were indestructible.

Plus two other minor changes: 

* Decos' servants are changed to New Attnam team rather than Decos' team (doesn't make sense for them to side with him once the revolution starts)

* The "Are you sure you want to wish for ___?" prompt is now restricted to Wizard Mode, as it could easily be abused to get ommel liquids and the like. Wishing in a normal game will be the same as in 0.50 now.

### DIFF
--- a/Main/Source/gear.cpp
+++ b/Main/Source/gear.cpp
@@ -357,7 +357,7 @@ void meleeweapon::AddInventoryEntry(ccharacter* Viewer, festring& Entry, int, tr
     Entry << " [" << GetWeight() << "g, DAM " << GetBaseMinDamage() << '-' << GetBaseMaxDamage();
     Entry << ", " << GetBaseToHitValueDescription();
 
-    if(!IsBroken() && !IsWhip())
+    if(!IsBroken())
       Entry << ", " << GetStrengthValueDescription();
 
     int CWeaponSkillLevel = Viewer->GetCWeaponSkillLevel(this);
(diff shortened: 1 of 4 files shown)


rvanherel:
Awesome changes Pent. Can we merge these into the development branch?

Clawzfire:
Yeah. I'll submit pulls for future changes right to devel so we can keep `master` stable, but this batch I've actually been using for the past 6-7 months (never got around to getting it into `master` due to a slew of excuses I could come up with), so I'd consider it thoroughly playtested.

Alternatively we could make `master` the working space and create branches as 'snapshots' of stable points/releases, which may be cleaner than what we have currently. Since the version numbers are pretty arbitrary we could just call any point at which it's stable or has some extra features 0.51.1 or 0.51-1 or whatever to distinguish it from older revisions.

rvanherel:
Hmm. I think you're right. Pushing changes directly to 'master' is more clean and logical. I don't think we need to use such an expansive parallel development-branch workflow. I think that instability can be contained pretty efficiently with one development trunk and anyhow, coders may find it confusing to walk into a project with abstruse workflow. 
I like the idea of using branches as release snapshots. Shall we merge the dev branch into 'master' then?
No need to make excuses :) we are all doing it for the love.

andrewtweber:
I am fine with having master be the working space, but we should use tags for releases, not branches.

So:
**Attnam/master** - working space
**rvanherel/master** - fork of Attnam/master. Submit pull requests to Attnam/master. Should mostly be used for small changes
**rvanherel/FireSystem** - branch of rvanherel/master, for large feature additions. When feature is mostly complete, should first be merged into rvanherel/master, and then submit a pull request to Attnam/master.

Then we can do tag releases of Attnam/master as snapshots. A tag is exactly what we want, it's just a stopping point saying all commits up to this point are part of 0.51, etc.

PR #35, opened 2015-04-19
Modernized the IVAN build system, and cleaned up user-specific data from repo

emlai:
__Summary:__
* Updated the Autotools configuration files (`configure.ac` and `Makefile.am`) to get rid of deprecated stuff.
* Removed some user-specific build configuration files and objects files from the repo. Still left the release-specific configuration files (e.g. `Makefile.in` files, `configure`, etc.) lying around. Should they be left in the repo, or included only in release tarballs, as is the common GNU convention?
* Added the `autogen.sh` script for easy generation of release-specific config files.
* Added .gitignore to help keep repository clean of user-specific build configuration files, build products, etc.
* Deleted those ignored files from the repo.
* Fixed an undefined reference error when compiling without wizard mode.


rvanherel:
Awesome! It compiles :)

andrewtweber:
Awesome work @rvanherel !

PR #36, opened 2015-06-17
SDL2 Support

murlock:
Please found a SDL2 support for Ivan.

SDL1.2 is still supported and will be used as fallback if SDL2 is not detected.

What is missing : 
- Test on BIG_ENDIAN computer because I left empty function for SDL2 (will try to test with my PI2 later)
- Complete test of keyboard (low level event (backspace, escape, ...) and high level event are managed differently : check https://wiki.libsdl.org/MigrationGuide#Input)

### DIFF
--- a/FeLib/Include/graphics.h
+++ b/FeLib/Include/graphics.h
@@ -48,12 +48,15 @@ class graphics
   static void SetSwitchModeHandler(void (*What)())
   { SwitchModeHandler = What; }
 #ifdef USE_SDL
-  //static SDL_Surface* Screen;
+#if SDL_MAJOR_VERSION == 1
+  static SDL_Surface* Screen;
+#if SDL_BYTEORDER == SDL_BIG_ENDIAN
+  static SDL_Surface* TempSurface;
+#endif
+#else
   static SDL_Window* Window;
   static SDL_Renderer *Renderer;
   static SDL_Texture *Texture;
-#if SDL_BYTEORDER == SDL_BIG_ENDIAN
-  static SDL_Surface* TempSurface;
 #endif
 #endif
  private:
--- a/FeLib/Include/graphics.h
+++ b/FeLib/Include/graphics.h
@@ -47,14 +47,17 @@ class graphics
   static rawbitmap* GetDefaultFont() { return DefaultFont; }
   static void SetSwitchModeHandler(void (*What)())
   { SwitchModeHandler = What; }
- private:
-  static void (*SwitchModeHandler)();
 #ifdef USE_SDL
-  static SDL_Surface* Screen;
+  //static SDL_Surface* Screen;
+  static SDL_Window* Window;
+  static SDL_Renderer *Renderer;
+  static SDL_Texture *Texture;
 #if SDL_BYTEORDER == SDL_BIG_ENDIAN
   static SDL_Surface* TempSurface;
 #endif
 #endif
+ private:
+  static void (*SwitchModeHandler)();
 #ifdef __DJGPP__
   static ulong BufferSize;
   static ushort ScreenSelector;
--- a/FeLib/Source/graphics.cpp
+++ b/FeLib/Source/graphics.cpp
@@ -29,7 +29,7 @@
 void (*graphics::SwitchModeHandler)();
 
 #ifdef USE_SDL
-#ifdef SDL_MAJOR_VERSION == 1
+#if SDL_MAJOR_VERSION == 1
 SDL_Surface* graphics::Screen;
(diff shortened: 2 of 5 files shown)


emlai:
Awesome! I tested it and noticed a few things:

* Fullscreen mode on my system now makes the game run blazingly fast, so fast that all actions that take time (rest, go, etc.) feel almost instantaneous.

 To prevent this, we could add a `SDL_Delay(10)` inside the graphics rendering loop, but I'm not sure if this is the best solution. Any ideas?

* The game is now stretched to fit the entire screen (instead of letterboxing), e.g. with my display the width of the game tiles is greater than their height, which looks weird. Also by default SDL2 uses nearest neighbor scaling which doesn't look so good either. There's also some weird glitch in the menu when I launch the game. I took some screenshots to demonstrate these, you can find them [here](https://www.dropbox.com/sh/ewmnycvqigbqh0p/AAAirDcbYuGEUEGqszb5XMh_a?dl=0).

 I then tried using `SDL_WINDOW_FULLSCREEN` (the "real" fullscreen mode, [according to the SDL wiki](https://wiki.libsdl.org/SDL_SetWindowFullscreen)) instead of `SDL_WINDOW_FULLSCREEN_DESKTOP` for the fullscreen mode, and all of the above problems went away. I also use `SDL_WINDOW_FULLSCREEN` in one of my own SDL2 projects and prefer it to `SDL_WINDOW_FULLSCREEN_DESKTOP`. I think it produces nicer results. I made a [patch](https://github.com/emlai/ivan/commit/076ba12faf17d93152a79f7a5080457458ee78f2) with the changes.

murlock:
I'll do more test more in fullscreen for speed issue, but I think SDL_Delay is required.

Regarding SDL_WINDOW_FULLSCREEN_DESKTOP, I'll test with SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, "linear") for quality.
For ratio issue, I'll give a look in SDL API, I may have missed something.

Regarding issue with full screen, I've met same issue when switching windowed/full screen and I tough that calling BlitDBToScreen will be ok. 

Otherwise, it is ok for your patch.

emlai:
@murlock Found a fix for the ratio issue: `SDL_RenderSetLogicalSize`.

emlai:
I also noticed that if we add `SDL_WINDOW_ALLOW_HIGHDPI` to the window flags in SDL2, `SDL_WINDOW_FULLSCREEN_DESKTOP` looks good on my display, with no need to blur the pixels with `SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, "linear")`. So I guess we won't need `SDL_WINDOW_FULLSCREEN` after all.

I made the changes in my branch, [diff here](https://github.com/murlock/ivan/compare/master...emlai:master). Now we only need to fix the menu glitch and the speed-up.

murlock:
Great :)

emlai:
Found the reason why the window was not being redrawn correctly when switching between windowed/fullscreen: apparently `SDL_WINDOWEVENT_RESIZED` needs to be handled on OS X. So that's fixed.

There's still a small related bug that shows a frame of the menu in its windowed size (as shown in my screenshots) for a split second just before rendering the fullscreen sized view correctly.
Also for some reason the fade-in effect doesn't show in fullscreen (in windowed mode it works correctly though). My guess is that there might be a superfluous `BlitDBToScreen` somewhere but I couldn't find anything yet. But this is quite minor, and could IMO also be fixed later.

I also found a couple of other minor bugs while bug-hunting. A patch for all of these can be found in the same diff I linked above.

emlai:
I realized that the scaled resolution with nearest neighbor interpolation only produced good results for me because my screen height is a multiple of IVAN's hardcoded screen height. So every in-game pixel occupies exactly 3x3 actual pixels on my screen.
Of course, for displays whose height is not a multiple of 600, using nearest neighbor interpolation would produce warped results.

To solve this, we could do a quick workaround: using linear interpolation as a fallback on displays whose height is not a multiple of 600 or width not a multiple of 800.
Maybe at some point we can then make something clever to scale and letterbox the game appropriately to get nice sharp unwarped pixel art graphics at all resolutions >= 800x600, if we want.

I wrote [some code](https://github.com/emlai/ivan/commit/8ff87219d465c14598d5270ebcbd895d751d5b85) to enable either nearest neighbor or linear interpolation based on the current display resolution. It should work, let me know if it doesn't.

rvanherel:
Mugs like me might like to be aware of this glitch when using MinGW and gcc while implementing SDL2:
http://stackoverflow.com/questions/22446008/winapifamily-h-no-such-file-or-directory-when-compiling-sdl-in-codeblocks

PR #43, opened 2015-06-25
Fix warnings

emlai:
These commits silence most warnings emitted by compiling with `-Wall`, `-Wextra`, `-Wold-style-cast`, `-pedantic`, etc.

There are still many warnings about “dangling else” which should probably be fixed, but I wasn’t entirely sure so I left them out for now.

Other warnings that may still need fixing:
- “delete called on class that has virtual functions / is abstract but has non-virtual destructor”
- a bunch of member function hiding virtual functions
- a bunch of unused variables

Commit `415af4e` also makes the code compilable as C++11.


andrewtweber:
Awesome, awesome work!

PR #49, opened 2015-07-08
Make some string questions escapable

### LINKED ISSUE #48: Make all StringQuestions escapable?
You can cancel reading all other kinds of (non-instantaneous) scrolls, so it would be nice to be able to do so with scrolls of detect material as well. Or is there a specific reason why this is not allowed?

emlai:
Prompts that are affected:
- Scroll of change material
- Scroll of detect material
- Scroll of wishing
- Polymorph with polycontrol
- Summoning monsters in wizard mode.
- Naming pets
- Engraving

Non-scroll wishes are unaffected.

Resolves #48.

### DIFF
--- a/Main/Include/game.h
+++ b/Main/Include/game.h
@@ -243,7 +243,7 @@ class game
   static int GetTeams() { return Teams; }
   static void Hostility(team*, team*);
   static void CreateTeams();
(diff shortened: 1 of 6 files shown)


andrewtweber:
You are fast o_O

PR #62, opened 2015-07-24
Some minor fixes

emlai:


andrewtweber:
What was the point of all those `int esko = esko = 2` lines?

emlai:
@andrewtweber Haha, I was wondering that too. It turns out you can use them for conditional breakpoints. Just place the breakpoint to the `esko` line and it will only activate if the `if` is entered.

The reason that a simple `;` or `int esko = 2;` can't be used is that in some IDEs it won't work for some reason — the breakpoint will trigger after the `if` — probably because the `if` is optimized out or something. Initializing the dummy variable with itself prevents this.

I can't think of another explanation for them.

andrewtweber:
Ha, interesting. That makes sense

PR #66, opened 2015-08-14
Removes `GetDrawDisplacement()` for kamikaze dwarves and simply modifies …

rvanherel:
…the leg, groin and arm bitmaps in Humanoid.pcx to match how they were originally intended to look.


emlai:
Awesome! This is probably the right solution and it seems to work so I merged it. I also got rid of `humanoid::GetDrawDisplacement` since it was there just to make that kamikaze dwarf bodypart displacement thing possible and is thus not needed anymore.

PR #67, opened 2015-08-14
C++11 migration, part 1

emlai:
These commits slightly improve readability (especially in the case of range-`for`), maybe efficiency in some areas, and of course, make the source code more "modern" and slightly more attractive to C++ developers.

When I have more time I will do some more (drastic) changes, e.g. smart pointers (nobody wants raw `new`/`delete` in their C++ anymore), move semantics, auto, lambdas, etc.

Btw was there some specific reason why `truth` was typedefed to `long`? I can't think of any, and changing it to `bool` made save/bone files incompatible. We could of course still read/write bools as longs from/to files to stay compatible with saves/bones, but is it worth it, what do you think?

### DIFF
--- a/FeLib/Include/bitmap.h
+++ b/FeLib/Include/bitmap.h
@@ -26,9 +26,9 @@ typedef void (*bitmapeditor)(bitmap*, truth);
 struct blitdata
 {
   bitmap* Bitmap;
-  podv2 Src;
-  podv2 Dest;
-  podv2 Border;
+  v2 Src;
+  v2 Dest;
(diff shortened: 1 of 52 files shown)


rvanherel:
Haha, this means my compiler is out of date :banana:

emlai:
Hah yeah I always forget to `-DLINUX` too (even though I'm on OS X — should probably be fixed…).

rvanherel:
I have to set -DWIN32 :(
But it compiles with gcc4.8.1 !!

PR #72, opened 2015-10-08
Code formatting fixes

emlai:
This is mainly cosmetic changes (though a few do actually fix some warnings):
- The tab character was used for 4 levels of indentation instead of 8 spaces. This caused the indentation to render incorrectly in editors/diffs where the tab size was set to something else than 8. Now all indentation in the code uses spaces so everything is indented correctly regardless of the editor's tab size setting.
- There were some huge lines of code in some places, some of which were well over ~120 columns. These are now split so that the code can more easily fit in one editor/diff window while being formatted in a readable way at the same time.
- There were some inconsistencies in brace/indentation/whitespace styles. These are now fixed in favor of the majority of the code. Idea: should we set up a config file for an automatic code formatting tool like clang-format or uncrustify, so that everyone could easily make sure that their contributions conform to the official IVAN style before committing?


emlai:
Merged in 351b4c0530e84c5aa4d9f026aefd95fae664a5d3.

andrewtweber:
Cool, I didn't want to mess things up trying to resolve those conflicts so thanks for taking care of it and merging in.

Issue #17, opened 2015-03-02
Lanterns causing fires


fejoa:
Smashing lanterns should be a source of fire initiation

AlexMooney:
Are we calling this one done, or, if not, can we add a list of stuff we want to add to finish this?

fejoa:
Yeah, it's pretty much done. I think the fire subsystem is pretty well fleshed out. There are still bugs to catch and there are other subsystems to keep working on. If need be, the fire subsystem can be revisited at a later date.

PR #81, opened 2016-02-14
WIP: Lanterns cause fire

AlexMooney:
https://github.com/Attnam/ivan/issues/17

This makes lanterns deal fire damage to objects, terrain, and creatures
present in the square the break in.

There was a seg fault throwing a lantern at genetrix so I threw in lots
of debug printing to make sure this functions. Also, 100 fire damage is
almost certainly too much. Move it to a variable on Lantern later.

Perhaps redo this to use an explosion and change explosions to have
option to do only FIRE damage?

### DIFF
--- a/Main/Include/level.h
+++ b/Main/Include/level.h
@@ -91,6 +91,7 @@ struct explosion
   int RadiusSquare;
   int Size;
   truth HurtNeutrals;
+  truth FireOnly;
 };
 
 struct beamdata
@@ -175,7 +176,7 @@ class level : public area
   room* GetRoom(int) const;
   void SetRoom(int, room*);
   void AddRoom(room*);
-  void Explosion(character*, cfestring&, v2, int, truth = true);
+  void Explosion(character*, cfestring&, v2, int, truth = true, truth = false);
   truth CollectCreatures(charactervector&, character*, truth);
   void ApplyLSquareScript(const squarescript*);
   virtual void Draw(truth) const;
--- a/Main/Include/miscitem.h
+++ b/Main/Include/miscitem.h
@@ -92,6 +92,7 @@ ITEM(lantern, item)
   virtual truth AllowAlphaEverywhere() const { return true; }
   virtual int GetSpecialFlags() const;
   virtual truth IsLanternOnWall() const { return GetSquarePosition() != CENTER; }
+  virtual void Break(character*, int);
  protected:
   virtual int GetClassAnimationFrames() const { return !IsBroken() ? 32 : 1; }
   virtual col16 GetMaterialColorA(int) const;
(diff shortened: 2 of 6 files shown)


fejoa:
So, testing this at the moment. Some thoughts:

- Threw a lantern at a backpack full of gunpowder and crashed the game. Will check this out further.
- Noted that carnivorous plants do not burn ...yet #82
- Noticed that the damage caused by an exploding lantern is possibly excessive. There is a post explaining the [activation thresholds for some materials](https://attnam.com/posts/26162). I would recommend a fire damage of 5 be inflicted instead of 100. This would catch parchment, flesh, and plant fiber on fire, without affecting leather for example.

fejoa:
Ok so after a little source dive, I think maybe the function you need to implement is actually `Lantern::ReceiveDamage` not `Lantern::Break`. Inside `Lantern::ReceiveDamage` you can cause a little explosion. Have a look at [how this is implemented for a bottle full of vodka](https://github.com/Attnam/ivan/blob/master/Main/Source/miscitem.cpp#L2153). You could restrict the explosion to the case only where the lantern is thrown, but maybe you wish to allow more general behaviour like cascading explosions of lanterns due to another explosion.

AlexMooney:
Thanks for the look.  I've added explosions that only deal fire damage (but no new graphic for them).  Wands of fireball and lanterns use these new explosions.  I set the lantern explosion damage to just 5, which still seems to be plenty.  I wrote some code for `Lantern::ReceiveDamage`, but it was being overridden by `Lantern::Break`, so I just went back and fixed that into something workable.

As far as I can tell, this works and doesn't cause any problems, besides possibly balance related ones.  The good news is that it's very hard to hit most enemies with a thrown lantern.  The bad news is that even creatures a couple of squares away from the lanterns still take some damage.

It would be better if there was a flammable oil fluid that the lanterns spilled and ignited.  I somewhat doubt flaming fluids are already covered, but I don't really know.

fejoa:
Awesome! When I got up this morning, I realised I was mistaken and the explosion really needed to happen in `Lantern::Break`, otherwise it would try to make an explosion in the wrong place in the code. Looks like you already sorted it out. I'll try test this at the earliest possible moment.

AlexMooney:
Misremembered.  Not potions; it's in `item::Break`.
https://github.com/Attnam/ivan/blob/master/Main/Source/item.cpp#L816

fejoa:
I found a bug, but it only occurs in Wizard Mode when you have the lights up (3-key a couple of times) and you can see everything. It happens at [this point](https://github.com/AlexMooney/ivan/commit/fd3610facb67bbddc7bd1ffb95c40a91c734a7ca#diff-118e50709c4f22e80109b8ba2c7c826eR1607) in the code, but I suspect that it may be for complicated reasons (related to illuminances during the explosion and the fact that the lantern is also switching off its emitation at some point??).
The bug does not occur in survival mode with or without secondary light sources present.

AlexMooney:
It might be because drawing the explosion invokes `LuminanceMaskedBlit` and the lighting has not been updated yet.  We could try moving the explosion to before the lantern is swapped with a broken lantern and see how that works.  Hopefully it doesn't cause some kind of recursive lantern breaking stack overflow nonsense...

fejoa:
Merged in the interests of progress. The wizmode exploding lantern crash still stands though o.0

PR #86, opened 2016-02-24
[WIP] More god effects

AlexMooney:
Generally give gods more things to do and make an attempt to balance them against one another.

### DIFF
--- a/Main/Source/gods.cpp
+++ b/Main/Source/gods.cpp
@@ -964,23 +964,34 @@ void nefas::PrayGoodEffect()
       }
   }
 
-  mistress* Mistress = mistress::Spawn(RAND() & 7 ? 0 : TORTURING_CHIEF);
-  v2 Where = game::GetCurrentLevel()->GetNearestFreeSquare(Mistress, PLAYER->GetPos());
+  if((GetRelation() > 200) && RAND_N(5)) {
+    int Chief = 3000/GetRelation();
 
-  if(Where == ERROR_V2)
-  {
-    if(PLAYER->CanHear())
-      ADD_MESSAGE("You hear a strange scream from somewhere beneath.");
+    mistress* Mistress = mistress::Spawn(RAND_N(Chief) ? 0 : TORTURING_CHIEF);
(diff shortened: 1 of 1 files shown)


andrewtweber:
Could you explain how the charisma training works? Does it, like the weapon enchantment, get less likely the higher your CHA gets? 

Of course, CHA is pretty useless except for sci talking to Richel Decos (I don't think the shop prices are that useful since you always end up with more money than you need), but if we ever made it more useful I want to make sure Dulcis isn't too powerful.

AlexMooney:
Attribute training always gets slower the more of that attribute you have.
- https://github.com/Attnam/ivan/blob/master/Main/Source/char.cpp#L8113
- https://github.com/Attnam/ivan/blob/master/Main/Include/ivandef.h#L822

The function (for the player) is
`GainedExp = ((10 * 5e7) * (100 + Value) - 100 * (Current_Attribute * 5e7)) * Speed * 2e-8`
Where Value is the middle argument to `EditExperience` and `Speed` is the third argument.  Every 50 million Exp is 1 full attribute point.  For NPCs, speed is increased by 1.5x and the first "10" changes to their base attribute value for the attribute being trained.

I've modified the numbers such that she gives at least 0.5 points of Cha / prayer when Cha is 10 and at most 1.0 points of Cha / prayer when Cha is 15.  The effect strength is a function of your relationship to her.  She can raise Cha  up to at most 30, which will take roughly 75 prayers with her at maximum effect, starting at Cha 10.

For reference, praying will raise Wis to 30, but does so at only speed 1024, instead of 33k.

andrewtweber:
Awesome. I'll let someone else review the code itself but the ideas behind this seem good to me

fejoa:
I'm just going to merge this for momentum.

AlexMooney:
I'm not sure that really belongs in master.  We never discussed those
changes in Attnam; some people might object to changing Cruentis and
Legifer so drastically.  Maybe we should at least ask forgiveness..?

fejoa:
Sorry, I acted prematurely :)

PR #94, opened 2016-03-16
WIP: Add titanium int 25 in adamant material chain

AlexMooney:
`Density` is real, but `Strength` and `PriceModifier` are just guesses.

TODO:
- [ ] Add `TITANIUM` as config option for various metal items
- [ ] Review `PriceModifier` to make sure it's reasonable
  - In particular, prices are calculated proportional to `PriceModifier * Weight` so the lower density may undervalue it
  - More likely, it's too high; check other `IntelligenceRequirement = 25` materials for balance
- [ ] Review `Strength` to make sure it's reasonable
  - Make sure it's in line with other materials at this level of `IntelligenceRequirement`
- [ ] Add another metal with `IntelligenceRequirement = 30` (?)

### DIFF
--- a/Main/Include/confdef.h
+++ b/Main/Include/confdef.h
@@ -85,6 +85,7 @@
 #define DARK_GRASS (SOLID_ID + 64)
 #define LEAD (SOLID_ID + 65)
 #define BLACK_GRANITE (SOLID_ID + 66)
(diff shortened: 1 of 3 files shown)


AlexMooney:
Closing this in light of the discussion [here](https://attnam.com/topics/Titanium-and-other-metals)

PR #97, opened 2016-03-20
Permanent status always shows blue

AlexMooney:
Before this, permanent conditions would show as temporary before the
duration was reduced to the permanent value.


andrewtweber:
Wasn't there one - telecontrol or ESP - that never turned blue? Does this fix that one too?

AlexMooney:
If one of those states never turned blue, I don't think this will fix that.  This only makes it so that something that "will become" permanent, by having a higher value than the permanence constant, always shows as permanent.

PR #104, opened 2016-03-26
Player ghost

fejoa:
Some major changes: Ghosts generated from player characters now resemble the deceased.


emlai:
Hmm, I get [this error message](https://github.com/Attnam/ivan/blob/04e6fe51c1665336ea6655ce9f472e0be14ebdc7/Main/Source/database.cpp#L820) sometimes when I enter a level.

emlai:
I deleted all my bones and it works now so I guess `BONE_FILE_VERSION` just needs to be incremented.

PR #113, opened 2016-04-19
Cause the dungeon data file not to have to explicitly state the numbe…

fejoa:
…r of dungeons


emlai:
Why are the removed lines commented out rather than… well, removed? Or am I missing something?

Also why was `mDungeons` added even though it is not used? Or will it be used in some future changes?

fejoa:
You're right, this was hastily done. I will take this away and improve it.

PR #114, opened 2016-04-22
Auto count dungeons

fejoa:


Script/dungeon.dat
```diff
@@ -95,6 +94,8 @@ Team BETRAYED_TEAM;
   KillEvilness = 10;
 }
 
+/* Dungeons to be included in the game */
+
 Include "dungeons/Attnam.dat";
```
andrewtweber: Is it easy to include `dungeons/*.dat` so this file doesn't have to be updated?

emlai: Yes, but it adds some non-cross-platform code (or an external library dependency, e.g. boost::filesystem). I'll implement this quickly.

emlai: Done: #115

fejoa:
Thanks fellas for the diligent work, it is nice :)

PR #115, opened 2016-04-23
Auto-detect dungeon data files

emlai:
Removes the need to specify each dungeon data file separately in `dungeon.dat`.

Haven't tested on Windows but should work.


fejoa:
This compiles and runs sweet-as in windows by the way.

emlai:
@fejoa Great!

PR #116, opened 2016-04-25
World terrains to data files

fejoa:
Some changes to make wterrains behave more like lterrains
See commit messages for ee8b335 and 5f62086

Please check for cross-platform compatibility, and stability.


emlai:
Seems to work well!

PR #125, opened 2016-05-07
WTerrain data abstraction to data files

fejoa:
`NameStem` and `UsesLongArticle` have been put into gwterra.dat and owterra.dat
`AttachedDungeon` and `AttachedArea` appears in owterra.dat
`Priority` appears in gwterra.dat

Note that changing values like `Priority` are quite likely to radically break the game.


andrewtweber:
Very nice

emlai:
Great! Seems to work fine.

Issue #40, opened 2015-06-22
'g'o command does not work with numpad


andrewtweber:
On IVAN 0.50.3 Windows, when I press 'g' to go and then the arrow key to go in the direction, it simply moves the player 1 tile

rvanherel:
It's true. Try minimising the window, and then maximizing again? The problem goes away. But then it comes back if I chat to npc's or do something else. Odd.

emlai:
Weird. Is this with SDL2 or 1? And which version specifically?

rvanherel:
SDL2, from this windows build: https://github.com/Attnam/ivan/releases/tag/v0503

andrewtweber:
Yes, SDL2. For me the minimize/maximize trick doesn't work either. It happens in both fullscreen and windowed.

murlock:
I've tested with my Linux version, it works with my azerty keyboard, I still have to download 'official' 0.50.3 to test it.
I didn't test fullscreen / windowed mode.

emlai:
Unfortunately I can't reproduce this on OS X, not even when running the Windows executable through Wineskin. What flags was that executable compiled with?

rvanherel:
Mock me if you will, (and you probably will) but I use this fossil for compiling: mingw gcc version 3.4.2 
Using mingw32-make.exe I invoke 
<code>-B -f c:\ivan\ivanmgw-local.mak</code>

The makefile looks like the one below.

    CC = g++ -o

    FeLibDIR = C:\ivan\FeLib

    FeLibGCH =

    FeLibOBJ = $(FeLibDIR)/Source/bitmap.o $(FeLibDIR)/Source/config.o $(FeLibDIR)/Source/error.o $(FeLibDIR)/Source/feio.o $(FeLibDIR)/Source/felist.o $(FeLibDIR)/Source/femain.o $(FeLibDIR)/Source/femath.o $(FeLibDIR)/Source/festring.o $(FeLibDIR)/Source/fetime.o $(FeLibDIR)/Source/graphics.o $(FeLibDIR)/Source/hscore.o $(FeLibDIR)/Source/rawbit.o $(FeLibDIR)/Source/save.o $(FeLibDIR)/Source/whandler.o

    SDL2DIR = C:\MinGW\include\SDL2

    IVANDIR  = C:\ivan\Main

    IVANBIN  = IVAN.exe

    IVANGCH =

[...]

PR #152, opened 2016-08-07
CMake

### LINKED ISSUE #149: Convert build system to CMake?
The current Autotools build system is a bit old and clunky. A more modern solution such as CMake would provide many benefits. It's also what most C++ projects are using nowadays.

The main advantage is that CMake is not a build system, but a build system _generator_. It can generate Makefiles, various IDE projects, and other build systems. This is very convenient, since it allows everyone to work in a way that best suits their workflow.

Many more reasons for and against are already listed on the web, so I won't repeat them here. Here are a few links:

http://stackoverflow.com/a/18291580/3425536
http://stackoverflow.com/q/5837764/3425536
https://www.quora.com/Is-it-worth-replacing-autotools-with-cmake

What do you think?

(I could do this.)

emlai:
Fixes #149.

~~WIP, don't merge yet. Needs testing on Windows and Linux.~~

To generate a build system, run `cmake .`.

I don't know which build system CMake generates by default on Windows, but you can explicitly specify one if you want, e.g. `cmake -G "MinGW Makefiles" .` or `cmake -G "Visual Studio 14 2015" .`. `cmake --help` lists all possibilities.

After that just use the generated build system like you normally would, e.g. `make && make install`.

~~Can someone confirm that this works on Windows? Also, is the icon file bundled correctly on Windows (commit 3d6d114)?~~

TODO:
- [x] Fill in the build instructions section in `INSTALL` for Windows.
- [x] Merge #159.
- [x] Rebase this pull request on top of the new master and make necessary changes to convert the new build system stuff introduced in #159 to CMake.
- [x] Test that the rebase didn't break anything.
- [x] Fix banana pick-axe terminal on MinGW.
- [x] Add RtMidi flags for Linux.
- [x] Test on Linux.


AlexMooney:
Tried to test this on linux, but I got this old graphical bug:
image

Also, this version is trying to save the scores in `/usr/local/var/ivan/ivan-highscore.scores` (the directory was not automatically created) instead of `/var/lib/ivan/ivan-highscore.scores`.

emlai:
I think that graphics bug is unrelated to CMake, I've had it before. Or is it consistently reproducible now?

Regarding `/usr/local/var` vs `/var/lib`, I think `/usr/local/var` is the right one:

1. The GNU documentation [here](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html) says that the `localstatedir` "should normally be `/usr/local/var`".
2. The [CMake module](https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html) that defines these GNU install directories gives `/usr/local/var` when requesting `LOCALSTATEDIR`.

AlexMooney:
I just attempted to reproduce the bug, but it didn't work. Sorry. :disappointed:

emlai:
I made an issue for that bug: #154

fejoa:
So I used a cmake gui to create a makefile in windows:

win_cmake_settings

... and got the following error:

win_compile

It gets most of the way there, I think it's a small step to fixing, but I can't immediately see what it might be. I have SDL2 libraries (2.0.4) pasted into my MinGW distro (why, oh why?), but cmake found all the SDL2 stuff seemingly. What's this about `pthread.h - not found`?
Any ideas on what I can try next?

emlai:
I'm able to reproduce it:

```
[100%] Linking CXX executable ivan.exe
C:/MinGW/SDL2-2.0.4/lib/x86/SDL2main.lib(./Release/SDL_windows_main.obj):(.text[_main]+0x0): multipl
e definition of `main'
c:/mingw/bin/../lib/gcc/mingw32/5.3.0/../../../libmingw32.a(main.o):(.text.startup+0x0): first defin
ed here
Warning: corrupt .drectve at end of def file
C:/MinGW/SDL2-2.0.4/lib/x86/SDL2main.lib(./Release/SDL_windows_main.obj):(.text[_main]+0x12): undefi
ned reference to `SDL_main'
collect2.exe: error: ld returned 1 exit status
Main\CMakeFiles\ivan.dir\build.make:765: recipe for target 'Main/ivan.exe' failed
mingw32-make[2]: *** [Main/ivan.exe] Error 1
CMakeFiles\Makefile2:139: recipe for target 'Main/CMakeFiles/ivan.dir/all' failed
mingw32-make[1]: *** [Main/CMakeFiles/ivan.dir/all] Error 2
Makefile:126: recipe for target 'all' failed
mingw32-make: *** [all] Error 2
```

But the errors are a bit different. For example, I get a "multiple definitions of `main`" error.

emlai:
Weird, if I just add `#undef main` after `#include "SDL.h"` in `femain.cpp`, then it works.

fejoa:
I can confirm it now compiles. I added `#undef main` after `#include "SDL.h"` as you suggested.

I have a funny little windows terminal that pops up when I open IVAN.exe. The terminal window even has the banana pickaxe logo :banana: 

My executable also installed under the `\ivan\Main` directory. Ideally it should install to the `\ivan` directory.

The executable size is 8 MB. What's the optimization setting? I use `-O0` which gives me around 4 MB these days.

fejoa:
You probably won't like this: The program crashes when you try to exit. Is that related to the funny change to `femain.cpp`?

emlai:
> My executable also installed under the `\ivan\Main` directory. Ideally it should install to the `\ivan` directory.

Yeah, this is being [worked on](https://github.com/emlai/ivan/pull/3/commits/2bddf0135081c861c08d92012b44b90c51e4844a). I'll merge that to this PR soon.

> The program crashes when you try to exit.

Yeah, that's probably caused by using plain `main` instead of `SDL_main`. So `#undef main` most probably isn't the right thing to do. The "undefined reference to `SDL_main`" error sounds like libSDLmain isn't being linked in. I'll investigate soon.

emlai:
By the way, #159 should be merged before this one because it modifies the build system and those changes must be converted to CMake. (I put this as a to-do at the top.)

fejoa:
Ok I merged #159 since it is pretty much in its final form, as far as required changes to the build system is concerned.

fejoa:
Great stuff, the progam no longer crashes on exit.
Banana pick-axe terminal is still there.

emlai:
Rebased this on top of the new master, so the audio stuff should now work with CMake. Can you confirm that it still works for you?

emlai:
@fejoa Fixed the banana pick-axe terminal in commit 04757df902fa0637fbf615f5289c3d18a1197e8c.

fejoa:
Cool, it compiles and runs as normal. 
However, I can't find any MIDI devices in the config menu, and there is a command window in the background with the error: MidiOutDummy: This class provides no functionality.

fejoa:
Oh shivers. @emlai Maybe you should take a look at this makefile I use. It has some extra linker flags...
[ivanmgw-0507.zip](https://github.com/Attnam/ivan/files/474432/ivanmgw-0507.zip)

fejoa:
Ssy, #40 is fixed :runner:

emlai:
@fejoa Audio should now work on Windows: 905e407c660f1eac50524b08ae671a4369eb93be

We probably need to add some flags for Linux too. [The RtMidi doc](https://www.music.mcgill.ca/~gary/rtmidi/index.html#compiling) says both ALSA and JACK work on Linux. Which one should we use?

fejoa:
I'm pretty sure it's ALSA.
I think some stuff was integrated into an eclipse IDE build tool: https://github.com/Attnam/ivan/commit/c1a99a858c989f1d05975a17aea9c455701a2a7e

emlai:
Okay, it should probably work on Linux now. Need testing though.

fejoa:
It works! No terminal window :) Only I had to manually add `-mwindows` to the linklibs.rsp files.

emlai:
> Only I had to manually add `-mwindows` to the linklibs.rsp files.

Hmm, it should work out of the box if you generate a build system with cmake, since I added it to the link flags [here](https://github.com/Attnam/ivan/commit/04757df902fa0637fbf615f5289c3d18a1197e8c#diff-79a16ca3bc6e8579a24074d0faf2d3e3R129). It works for me at least.

fejoa:
Aha. I had to delete the CMake cache in the gui. Problem was at my end.
It works perfectly. Can merge this as far as windows is concerned.

emlai:
👍

emlai:
Tested on Linux, and it works! Haven't tested audio yet though.

fejoa:
Sorry bro, I was sick last couple of days :(

emlai:
NP, hope you're feeling better :)

andrewtweber:
Nice work @emlai. Just compiled smoothly on OSX and seems much cleaner than the old build method

emlai:
@andrewtweber Thanks. Agreed. Fun fact: now you can just do `make install` instead of `sudo make install`. 💪

PR #153, opened 2016-08-07
Don't force "the" article for holy books

### LINKED ISSUE #144: "The" holy book
In the current master branch, holy books are always referred to using the definite article "the". I believe using the indefinite article "a" would be more appropriate since there can be many holy books for each god. Some examples:

screen shot 2016-06-26 at 13 05 05

screen shot 2016-06-26 at 12 59 11

emlai:
Fixes #144.

### DIFF
--- a/Script/item.dat
+++ b/Script/item.dat
@@ -1705,7 +1705,6 @@ holybook
   FormModifier = 20;
   Adjective = "holy";
   NameSingular = "book";
-  ArticleMode = FORCE_THE;
   MainMaterialConfig == PARCHMENT;
   Price = 200;
   Roundness = 70;
@@ -3844,4 +3843,3 @@ firstbornchild
     NameSingular = "daughter";
   }
 }
-


Script/item.dat
```diff
@@ -3844,4 +3843,3 @@ firstbornchild
     NameSingular = "daughter";
   }
 }
-
```
AlexMooney: It's usually best to leave unrelated lines alone.

AlexMooney:
Won't we still end up with "The 2 holy banana of Oily Orpiv"?  I think it would be better to work on the `FORCE_THE` logic.

AlexMooney:
Ah, never mind!  I see that the other `FORCE_THE` items don't seem to stack.  :+1:

Script/item.dat
emlai: Yeah I know, however that instance didn't add any significant noise in the diff or blame, so I decided it wasn't worth it to leave it alone :)

PR #162, opened 2016-09-11
Nerf the perma stat gain from blink dogs and floating eyes a bit

fejoa:
Eat 15 blink dogs instead of three.
Eat fvey floating eyes instead of three.

Fixes #162.


fejoa:
#163

emlai:
Is this ready to merge? I think the two commits can be squashed to one.

fejoa:
Yep, go ahead :)

PR #164, opened 2016-09-11
Dwarfs are short again, and they don't have that missing bit in the m…

### LINKED ISSUE #143: Dwarves missing part of their body
In the current master branch, kamikaze dwarves look like this:

screen shot 2016-06-26 at 12 52 26

There's something missing below their belt.

Might be related to #66.

fejoa:
…iddle. But now they appear to levitate.

Fixes #143


emlai:
> But now they appear to levitate.

Hmm, would moving all the dwarf graphics down by 1 pixel help? That would also make them seem a little shorter than humans:

Before:
before

After:
after

fejoa:
I don't know if this is possible straight out of the box. The reason is that certain parts of the 16x16 tile are reserved for certain bodyparts, so things like groins, left legs, right legs torsos arms and heads only appear in pre-defined places. 
The devs did that `GetDrawDisplacement()` function as a workaround to put the dwarf on the ground I guess, but that ended up being the cause of the kamikaze bug.
We can put it on the list, but for now we might have to live with the inelegance.

emlai:
Yeah, this is good enough for now.

PR #184, opened 2016-09-20
Fix the imprisoned necromancer's lines

red-kangaroo:
One sentence didn't make much sense and several articles were missing in my opinion; plus 10000 > 10 000 because otherwise it's hard to read for me...


Main/Source/human.cpp
```diff
@@ -4640,7 +4640,7 @@ void necromancer::BeTalkedTo()
                                  "and possesses a cloak of unimaginable usefulness: The Shadow Veil.\""));
 
         game::TextScreen(CONST_S("The necromancer suddenly looks at you intently.\n"
-                                 "\"Tell you what, I think you can help me out. But first I'll need proof of your abilities.\n"
```
andrewtweber: "need proof" is actually correct

Main/Source/human.cpp
```diff
@@ -4613,9 +4613,9 @@ void necromancer::BeTalkedTo()
                                  "The necromancer scans the page from left to right several times. His face contorts:\n"
                                  "\"Bah! A canticle of Saint Petrus the Lion-Hearted!\"\n"
                                  "He continues down the page. His eye's widen:\n"
```
andrewtweber: This could also be fixed, should be "eyes" without an apostrophe

Main/Source/human.cpp
```diff
@@ -4613,9 +4613,9 @@ void necromancer::BeTalkedTo()
...
-                                 "\"O ho! 10000 bananas? It sounds bad out in the colonies. I'm sorry to hear about it.\"\n"
+                                 "\"O ho! 10 000 bananas? It sounds bad out in the colonies. I'm sorry to hear about it.\"\n"
                                  "\n"
-                                 "The necromancer allows the scroll to burn the ashes wither away in his hands.\n"
+                                 "The necromancer allows the scroll to to ashes in his hands.\n"
```
andrewtweber: This doesn't seem right. I like the suggestion in the issue, "... scroll to burn and the ashes wither away..."

Main/Source/human.cpp
red-kangaroo: Damn you, grammar!

Issue #128, opened 2016-05-23
Priestess of Silva doesn't read the news


serin-delaunay:
If you liberate Tweraif, almost all of the inhabitants change their dialogue to new lines reflecting their new freedom. The priestess of Silva still.uses the same lines as before, and those lines specifically refer to conditions under Decos, in the present tense. Her door is mostly kept open, so she should probably be able to see what has happened.

PR #192, opened 2016-09-29
Update priestess's friendly replies after Tweraif is freed

andrewtweber:
Addresses #128 

I'm not much of a writer so I'm fine if anyone has suggestions for her replies. (Also, not 100% sure on the back story - was Richel Decos sent by Petrus?)


andrewtweber:
I hadn't tried compiling this.. looks like it introduced a bug because `priestess::BeTalkedTo`  is a redefinition of `priest::BeTalkedTo` :(

emlai:
You're right. It doesn't compile. I reverted it.

PR #195, opened 2016-09-29
Fixing priestess's replies after liberation

andrewtweber:
Tested this time.

Not sure why the "Files changed" is only showing 1. Must be a GitHub bug since the original commit was already merged in and later reverted. After this is merged need to make sure `Script/char.dat` is up to date with all of her replies

### DIFF
--- a/Main/Source/human.cpp
+++ b/Main/Source/human.cpp
@@ -948,7 +948,14 @@ void priest::BeTalkedTo()
                   "for that but I need %ldgp for the ritual materials first.\"", Price);
   }
 
-  humanoid::BeTalkedTo();
+  static long Said;
+
+  if(GetConfig() != SILVA)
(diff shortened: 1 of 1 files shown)


fejoa:
@andrewtweber looks like you'll need to merge the changes to char.dat

andrewtweber:
Thanks. Bizarre GitHub fail I think. Committed directly to master

AlexMooney:
@andrewtweber when a commit is merged then reverted, git still thinks it's merged.  There is no special way to know what was reverted and what wasn't.  Another way to think of it is that there is nothing special as far as git is concerned about a revert commit; it just *happens* to have the reverse of some other stuff.  Git won't merge the same commit twice, so they can get dropped because git knows it was merged but doesn't know about the revert.

PR #197, opened 2016-10-15
Ring of unchanging prevents polymorphing

AlexMooney:
Also prevents changing back to normal


emlai:
Was this discussed somewhere? This kind of duplicates the behavior of the polycontrol ring, since that can be used to prevent polymorphing too.

AlexMooney:
It's discussed [here](https://attnam.com/topics/polymorph-victory), and a similar item is mentioned [in the dev notes](https://github.com/Attnam/ivan/blob/df6b70cbead8f5b9ecd1fe7a8f772c31c37069c5/Doc/Obsolete/Item%20ideas%20II.txt#L28).  This item is different from polycontrol in that it completely prevents polymorph and changing back from polymorph.  Think of it as a more convenient alternative, although it may be very unbalanced because you can uncontrolled polymorph into something powerful and then put the ring on.

serin-delaunay:
I think it should have a higher teleport priority. Teleporting away the ring of unchanging while the player is polymorphed into something powerful could really ruin someone's day. :D

AlexMooney:
@serin-delaunay I set it to 1000, like the other high-value rings.  Strangely, the amulets don't have `TeleportPriority` values!  You'd think all of them represent a significant target for the enemies.

serin-delaunay:
Yeah, the life saving and resistance amulets probably merit a high TeleportPriority too. Not so sure about the ESP amulet.
Aren't non-pet NPC's amulets invisible to the player when looking at them? If amulets are "hidden" I wonder how much sense it makes for them to be targets of item teleportation when equipped.

fejoa:
@AlexMooney is this ready as well?

AlexMooney:
I recall a discussion on the board that ended up with people not liking
this ring. I think we were going to make it only generate as a  scabies
champion item. I'm also inclined to shorten the amount of time it gives you
in lock after the ring comes off.

AlexMooney:
This was effectively merged in https://github.com/Attnam/ivan/pull/272.  Closing.

PR #208, opened 2016-10-26
Disallow negative damage

### LINKED ISSUE #206: Negative damage weapons
Sometimes a terrible weapon will have damage printed as `-2--3`, which is at least a typographical problem, if not a gameplay one.  Armor seems to bottom out at 0, but may also go negative.
image

Problems:
- Negative damages should not happen.
- Apparently the left side of the range symbol is the one with the smaller *magnitude* instead of the smaller *value*
  - `2-3`, and `7-10` implies `-3--2`, so `-2--3` is strange here too.
- Ranges use '-' instead of '–', which could be okay, but doesn't make my inner typographer happy.

See https://github.com/Attnam/ivan/issues/204 for the bug about these terrible enchantments being generated in the first place.

emlai:
Fixes #206.

This makes the minimum damage range 0-1, because of the `+ 1` in [`GetBaseMaxDamage()`](https://github.com/Attnam/ivan/pull/208/files#diff-48edbb851d9811c2b0edf79b14d91060R47). Is this okay or should the minimum damage range be 0-0?

### DIFF
--- a/Main/Source/item.cpp
+++ b/Main/Source/item.cpp
@@ -42,7 +42,7 @@ truth itemdatabase::AllowRandomInstantiation() const { return !(Config & S_LOCK_
 item::item() : Slot(0), CloneMotherID(0), Fluid(0), LifeExpectancy(0), ItemFlags(0) { }
 truth item::IsOnGround() const { return Slot[0]->IsOnGround(); }
 truth item::IsSimilarTo(item* Item) const { return Item->GetType() == GetType() && Item->GetConfig() == GetConfig(); }
-double item::GetBaseDamage() const { return sqrt(5e-5 * GetWeaponStrength()) + GetDamageBonus(); }
+double item::GetBaseDamage() const { return Max(0., sqrt(5e-5 * GetWeaponStrength()) + GetDamageBonus()); }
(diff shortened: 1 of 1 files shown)


fejoa:
I like it that 0-1 poses at least some threat to the player.

Main/Source/item.cpp
```diff
@@ -42,7 +42,7 @@ truth itemdatabase::AllowRandomInstantiation() const { return !(Config & S_LOCK_
 item::item() : Slot(0), CloneMotherID(0), Fluid(0), LifeExpectancy(0), ItemFlags(0) { }
 truth item::IsOnGround() const { return Slot[0]->IsOnGround(); }
 truth item::IsSimilarTo(item* Item) const { return Item->GetType() == GetType() && Item->GetConfig() == GetConfig(); }
-double item::GetBaseDamage() const { return sqrt(5e-5 * GetWeaponStrength()) + GetDamageBonus(); }
+double item::GetBaseDamage() const { return Max(0., sqrt(5e-5 * GetWeaponStrength()) + GetDamageBonus()); }
```
AlexMooney: `0.0` plz? :)

emlai: @AlexMooney `0.` is the style used by the rest of the code.

PR #222, opened 2016-12-09
The grand stollen knife of Attnam

serin-delaunay:
Based on the [Grand Dresden Stollen Knife](https://en.wikipedia.org/wiki/Stollen#Dresden_Stollen_festival) ([images](https://www.google.com/search?site=&tbm=isch&q=grand+dresden+stollen+knife)), I added a grand stollen knife of Attnam to the game. It's a unique gold-and-silver two-handed scimitar with extra volume, extra damage, and even less accuracy. It's main purpose is to make the game print the phrase "unbelievably inaccurate".

If the player can get their hands on it, it's an absolutely terrible weapon. It sells for only about 400 gold. If the player SoCMs it into a light metal such as arcanite/mithril/octiron/valpurium, and enchants it to +5 or more, then it becomes more viable, but (particularly in the case of mithril and valpurium) it still needs high arm strength.

So it probably won't unbalance the game.


andrewtweber:
Very amusing :) And yeah I don't think it'll upset the balance at all considering how ineffective it is and that you'd most likely have to steal it

serin-delaunay:
The grand stolen knife of Attnam.

PR #247, opened 2017-04-14
Save game on exit when not deleting saves

### LINKED ISSUE #101: Infinite loop when loading wizard mode save
Steps to reproduce:
1. Start new game.
2. Activate wizard mode.
3. Enter any area.
4. Quit, but answer no to "Remove saves?".
5. Load the save that's named after your character (i.e. not the AutoSave), either via "Start game" or "Continue game".
6. The game starts looping infinitely in [this loop](https://github.com/Attnam/ivan/blob/2783eacef86185cdfe0bdbeb86ea0389641ea7db/Main/Source/game.cpp#L421), with `InWilderness = true`.

If you activate wizard mode _after_ entering an area, this doesn't happen.

fejoa:
Fixes #101

### DIFF
--- a/Main/Source/game.cpp
+++ b/Main/Source/game.cpp
@@ -1759,6 +1759,9 @@ v2 game::NameKeyHandler(v2 CursorPos, int Key)
 
 void game::End(festring DeathMessage, truth Permanently, truth AndGoToMenu)
 {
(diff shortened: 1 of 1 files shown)


emlai:
Nice!

Issue #134, opened 2016-06-09
change resolution?


L3gend:
Sorry if this is the wrong place to ask this, but is there anyway to possibly change the resolution or have bigger tiles? The tiles are really small for me, my eyesight isn't the best in the first place. Just wondering if there might be a way to increase the resolution x2 in the future? Or scale the tiles x2 maybe? Even an ascii mode which could be scaled would be nice if possible.

fejoa:
Here is a workaround for the full screen problem:
1) boot up IVAN
2) go into configurations and change your pet name
3) quit IVAN
4) In the IVAN folder on your file system, you will find a configuration file called ivan.cfg
5) Open ivan.cfg in a notepad
6) Change the line `FullScreenMode = 0;` to `FullScreenMode = 1;`
7) Save your change to ivan.cfg and boot IVAN up again. It will be in full screen mode

L3gend:
thanks. I actually already managed to get the fullscreen working by using the cfg file from 504 and copying it into the 506 folder. This post was actually about some way to make the tiles and text bigger on screen even in fullscreen mode. even in fullscreen, I have a tough time seeing everything on screen and distinguishing things. An optional ascii mode would make it a little easier to identify thing for those of us used to old school roguelikes. Otherwise, a possible bigger tile set/ a way to scale the tiles to be bigger would be really nice. I think simply x2 would be enough. For myself at least.

emlai:
Currently the codebase assumes that the resolution is always 800x600, which makes this quite nontrivial to change. OTOH adding support for simple 2x-scaled rendering should be quite easy.

fejoa:
Might have to change something in `SDL_RenderSetLogicalSize` via https://github.com/Attnam/ivan/blob/df6b70cbead8f5b9ecd1fe7a8f772c31c37069c5/Main/Source/igraph.cpp#L71

emlai:
There's [SDL_RenderSetScale](https://wiki.libsdl.org/SDL_RenderSetScale) for this. PR incoming…

PR #249, opened 2017-04-14
Add option to scale graphics by 2x

emlai:
Partly addresses #134.

### DIFF
--- a/FeLib/Source/graphics.cpp
+++ b/FeLib/Source/graphics.cpp
@@ -50,6 +50,7 @@ graphics::modeinfo graphics::ModeInfo;
 
 bitmap* graphics::DoubleBuffer;
 v2 graphics::Res;
+int graphics::Scale;
 int graphics::ColorDepth;
 rawbitmap* graphics::DefaultFont = 0;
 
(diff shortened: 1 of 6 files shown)


FeLib/Source/graphics.cpp
```diff
@@ -292,6 +294,23 @@ void graphics::BlitDBToScreen()
...
+#warning Graphics scaling not implemented for SDL v1
+#else
+  // Scale the window, maintaining its center position.
+  v2 WindowPos, OldSize, NewSize = Res * NewScale;
+  SDL_GetWindowPosition(Window, &WindowPos.X, &WindowPos.Y);
+  SDL_GetWindowSize(Window, &OldSize.X, &OldSize.Y);
+  WindowPos += (OldSize - NewSize) / 2;
+  SDL_SetWindowPosition(Window, WindowPos.X, WindowPos.Y);
+  SDL_SetWindowSize(Window, Res.X * NewScale, Res.Y * NewScale);
+  SDL_RenderSetScale(Renderer, NewScale, NewScale);
```
emlai: I noticed this works even without this line. I don't know if this is just platform-specific behavior or an SDL bug so I'll leave it.

PR #250, opened 2017-04-15
Fix crash when stopping digging after becoming hungry

### LINKED ISSUE #120: Crash on becoming very hungry while digging
If I'm hungry, start digging a wall, and become very hungry before finishing, the game asks me whether I want to continue. at this prompt, some keypresses cause the game to crash.

emlai:
Also, the player's old wielded item wasn't being wielded again when stopping digging this way. This fixes that as well.

Closes #120.


fejoa:
@emlai awesome work fixing this!

PR #262, opened 2017-05-14
Make sirens randomly spawned monsters

red-kangaroo:
Sirens are fully implemented in the code, including their special attacks, they only cannot be generated. This fixes that.

I also added several skin colors, because who doesn't like exotic skin colors?


fejoa:
@red-kangaroo this has got lots of extra file changes. Would you be able to merge your changes in from an up-to-date feature branch? If you get stuck, I can explain how to do this.

red-kangaroo:
@fejoa Definitely! Sorry, I forgot to update the branch. Everything should be up-to-date, fixed and finished now.

fejoa:
Awesome, thanks @red-kangaroo

andrewtweber:
Seems kind of funky having a bunch of colors (red, blue, green), and then a couple ethnicities (Caucasian, Hispanic). I don't think there's any "PC" issues, it's just weird. Also does Hispanic even exist in this world? Maybe Tweraifan would be better

andrewtweber:
I just noticed the Asian ones in there too. If nobody else thinks it's weird then we can keep it I guess

fejoa:
It's weird now you mention it. The config names appear only in the script don't they? Let's change these to neutral config names.

red-kangaroo:
@andrewtweber @fejoa It's kinda weird, but the names only appear in script. Players will only see "siren" for all of them. It's probably just a description for devs, because you can easily imagine what an asian siren would look like. :)

andrewtweber:
Got it. Let's not worry about it then

PR #264, opened 2017-06-04
Global hall of fame

fejoa:


andrewtweber:
Just tested and it looks good to me.

@emlai is Heroku free? Any chance it will ever go down? Seemed pretty slow to respond when I first refreshed the page, like maybe they don't keep the site running or something. I'd be willing to host this on my server if you want

emlai:
Yes Heroku has a free tier, which this is using. The free tier apps are put to sleep when unused and waking them up takes a while. You can definitely host this on your server, that would be nice!

However it is currently too easy to send a fake high-score to the server and have it accepted, and I feel like it should be made less easy before putting this into production.

andrewtweber:
Gotcha. Tough since the code for both the game and the server are open source.

emlai:
Yep… I don't think it should try to prevent all possible misuses. Just not submitting high-scores when the data files have been modified or if running a non-release-build would be a good start.

fejoa:
No rush from my corner ```:) ``` 
Can we wait until it's fully featured?

fejoa:
I've been tooling around with MSYS2 and it's a pretty complete build system for building windows binaries now, I was really surprised.
I can build the hi-score server version of IVAN, and it works really well. Only catch is I have to package all the DLLs with IVAN.exe, but this is fine as long as it's not missing anything.

As far as windows compatibility goes, I'm happy to build the binaries ```:)```

fejoa:
What if we use an IVAN quiz question, one of a hundred or so, randomly generated by the server, which appears in game and the user has to answer it to submit their score? It should prevent robots, but not weirdos who spam manually.

We could also use filters on the hi-score page to filter the scores by registered user, so even if the combined hi-score page gets vandalized by anonymous users, those scores can be filtered out.

fejoa:
What about using SHA1 checksums? Some stackoverflow artcles to this end:
https://stackoverflow.com/questions/1435953/how-can-i-pass-git-sha1-to-compiler-as-definition-using-cmake
https://stackoverflow.com/questions/1704907/how-can-i-get-my-c-code-to-automatically-print-out-its-git-version-hash

We could put the SHA1 source and datafile hashes in the code at compile time and the server could match them?

emlai:
SHA1 helps to prevent accidental fake submissions if you have modified the data files locally.

But it doesn’t help against intentional fake submissions, you can e.g. calculate the SHA1 and send an HTTP request to the server with that SHA1 and an arbitrary score (without even opening the game). And that works on a registered account too.

The only way to prevent this that I can think of is to record every keypress and send them to the server to replay and verify the playthrough with an authentic copy of the game. This would also allow us to watch replays of other people’s games which I think would be very cool. But it requires a lot of work, so:

What do you think, do we want a replay system? And do we want to first release an early-access/beta version of the high score server without the replay verification (only the SHA1 checking), and later add the replay verification when (if) we have implemented it? Or wait until the replay system is implemented and then release the high score server with it?

fejoa:
It does seem like a lot of effort. What components would we need to make a replay system?
How does the server verify a game if monster generation, to take an example, is a psuedorandom process? I guess the monster generation order would also be logged? Level layout and current state is already encapsulated in bonefiles and savefiles, so I suppose we could make use of those resources. 
There's always merit in "doing it once, doing it right", but there's no harm in doing it in stages in the way you mentioned. We can test the appetite for a global hiscore server from the player community using an unsecured pilot.
The other thing is that exploits might get accidentally merged in. There would definitely need to be a scrubbing tool available on the hi score server. No lack of reliable administrators on Attnam's forum though!

fejoa:
What if we get the application to make a screenshot of the final moment in the game, be it victory or defeat, and get the server to check whether it is authentic using some image processing? Could check the image for player stats or death message? It's kind of done on the Attnam forums this way at the moment.
We could get the server to send a code to be watermarked into the image as well?

Come to think of it, maybe we can quickly "savefile" the game and send that to the server so that the server loads the game and plays the player's last move? Could additionally verify the final screenshot by correlation, as an example. Would need to make sure it's not too memory intensive,

fejoa:
@andrewtweber I wonder if it would be easiest to secure this by hosting the hiscore app behind the forum server, such that folks need to log in to the forum to verify and post their high score?
In this way high scores remain invisible until they get claimed by a forum user. If bogus high scores surface, these can be moderated away in the normal way. 
Could something like this be workable?

andrewtweber:
That'll be tough since they are written in different languages and I don't really have a (good) public API for the forum

PR #266, opened 2017-06-04
Helmets give protection from sound attacks

fejoa:
Conventional helmets give some finite protection against sound attacks directed at a character's head.
Full helmets give seemingly 100% protection against point-blank Enner beast's attack when directed at character's head.
Note that special helmets give same protection as conventional helmets.
Broken helmets give no protection whatsoever.

### DIFF
--- a/Main/Include/char.h
+++ b/Main/Include/char.h
@@ -140,6 +140,7 @@ struct characterdatabase : public databasebase
   int PoisonResistance;
   int ElectricityResistance;
   int AcidResistance;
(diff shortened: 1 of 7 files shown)


andrewtweber:
The rest of your body can still get damage from Enner even with a full helmet right?

Even so I think 100% is too powerful. Maybe cap it at 50% or 75% and then later we could introduce a special helmet (or earmuffs? lol) that could offer full protection.

PR #270, opened 2017-07-06
ToX entry improvements

red-kangaroo:
This PR is for @fejoa. Take what you like. ;)

Houses on the entry level changed to obsidian and ebony, to be more in line with the rest of the level (obsidian/black granite) and to prevent ruby farming.

Some decorations were added, plus some more people - they don't say much (many I'll be able to come up with some interesting lines for them one day), but make the town feel cozy. ;)


fejoa:
Awesome changes!

PR #271, opened 2017-10-10
Use Travis APT addon to work around dependency installation failure

emlai:


andrewtweber:
Thanks! Mac builds are down but the Linux ones were fixed so I'm just going to merge and cross fingers for when Travis is back up

PR #272, opened 2017-10-12
Add new items

red-kangaroo:
This PR adds quite a few new items, mostly to add variability to what the player can pick for his equipment and to at least a bit reduce the number of equipment with no substitute. The goal is to give the player a chance to pick - having both a ring of invisibility and a cloak of invisibility is good because it gives the player a choice. Having only a belt of levitation is not good, because if you want levitation, you will always need this belt.

Discussion thread is here:
https://attnam.com/topics/Testing-build


Script/item.dat
```diff
@@ -823,6 +931,9 @@ holybanana /* materialcontainer->banana-> */
   CanBeMirrored = true; /* will create a peel */
   AllowEquip = true;
   IsValuable = true;
+  CanBeWished = false;
```
fejoa: Should be allowed to wish for one

Script/item.dat
```diff
@@ -902,7 +1015,7 @@ pickaxe /* meleeweapon-> */
   BitmapPos = 0, 96;
   FormModifier = 35;
   NameSingular = "pick-axe";
-  Alias == "pick-ax";
+  Alias = { 3, "pick-ax", "pickaxe", "pick axe"; }
```
fejoa: Cool, didn't know this was possible 👍

Script/item.dat
```diff
@@ -1182,6 +1327,40 @@ scrollofteleportation /* scroll-> */
...
+
+scrollofbodyswitch /* scroll-> */
+{
+  Possibility = 2;
+  Price = 3000;
+  PostFix = "of body switch";
+  CanBeWished = false;
+  IsPolymorphSpawnable = false;
+  CanBeCloned = false;
+  AttachedGod = INFUSCOR;
```
fejoa: Maybe Scabies?

fejoa:
Looking good so far @red-kangaroo

Script/item.dat
Azba: Me neither, we should make more extensive use of this to get around some of the weirder text-adventure-game nonsense you have to do to wish for certain items.

e.g. Do you want some boots of kicking? Sure, but don't use "boots" otherwise I have no idea what you mean. Tell me "boot of kicking" and I'll give you two!

Script/item.dat
red-kangaroo: I have forbidden wishing for all artifacts on the reasoning that they should be unique/rare items that you should be excited about finding and they should make you consider your choices and adapt to its use, rather than wishing for any one you like. This stands for artifact weapons and armor, IMHO, but should I make consumable artifacts wishable again? I'm not opposed to that change, but for consistency, maybe it should be all consumable artifacts (the banana, the saga, the monograph)?

Script/item.dat
red-kangaroo: Yes, I tried to cover the most likely wishes (immortality) and spellings people could use. ;) If you have any wishes that are missing, tell me.

Script/item.dat
red-kangaroo: IDK, this feels more like demonic possession/mind manipulation than transformation/mutation, so I'd say Infuscor.

Azba: Gonna agree with re-kangaroo, Infuscor is the goddess of wrong knowledge and vile magic, and bodysnatching is absolutely all kinds of wrong and very vile.

fejoa: Ok I agree now, thanks guys

red-kangaroo:
I think this PR is complete. Unless there will be things to fix, I won't be doing further changes.

Main/Source/char.cpp
```diff
@@ -266,6 +266,65 @@ statedata StateData[STATES] =
...
+    SECRET,
+    &character::PrintBeginSwimmingMessage,
+    &character::PrintEndSwimmingMessage,
+    &character::BeginSwimming, &character::EndSwimming,
+    0,
+    0,
+    0
+  }, {
+    "Detecting",
+    SECRET|RANDOMIZABLE&~(SRC_MUSHROOM|SRC_EVIL),
```
fejoa: Try making this line read
```SECRET|(RANDOMIZABLE&~(SRC_MUSHROOM|SRC_EVIL)),```
with the extra paranthesis to help Travis be nice.

Issue #284, opened 2017-11-10
Wishing aliases


andrewtweber:
As reported on the forum, wishing for "wand of polymorph" will grant you a **ring of polymorph** instead because of the `polymorph` alias

PR #289, opened 2017-11-13
Fix wishing aliases bug

andrewtweber:
Fixes issue #284 

If the exact name match can't be found or the alias is longer (i.e. more specific) than the exact match, then it will use the alias.

| Wishing for...            | Yields...                 |
|---------------------------|---------------------------|
| polymorph                 | ring of polymorph         |
| ring of polymorph         | ring of polymorph         |
| ring of polymorph control | ring of polymorph control |
| ring of polymorph lock    | ring of unchanging        |
| ring of unchanging        | ring of unchanging        |
| wand of polymorph         | wand of polymorph         |


andrewtweber:
If this is approved, please "Squash and merge"

fejoa:
Wishing for "polymorph" on its own should polymorph the character. A task for later. 
Thanks for the fix @andrewtweber

PR #290, opened 2017-11-13
Code of conduct

andrewtweber:
Please "Squash and merge"... I had to make a minor amendment

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,74 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, gender identity and expression, level of experience,
+nationality, personal appearance, race, religion, or sexual identity and
+orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment
+include:
+
+* Using welcoming and inclusive language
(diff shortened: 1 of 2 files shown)


andrewtweber:
@serin-delaunay apologies this took some time to add. It seems like the Contributor Covenant is a growing standard these days, and this is copied verbatim with my email added as a contact. Hope this addresses any concerns about behavior between contributors. This doesn't address anything about the content of the game (slavery, etc.) but I think for that we can continue to go on a case by case basis - if reviewers find something offensive then hopefully it can be discussed civilly and ideally a compromise can be reached to keep the game's irreverent tone without being offensive.

andrewtweber:
Merging but comments still welcome

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,74 @@
...
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project team at capristo@attnam.com. All
```
serin-delaunay: Two or more different report contacts should be given, in case one of them is the subject of a report. It should also be clear which github accounts correspond to these email addresses.

fejoa: Good point about the second contact person. I'd rather not be that person because I'm a bit slow, in many respects. What about @AlexMooney (Ighalli)? He's been around even since official development days.

PR #291, opened 2017-11-18
Skeleton puppy

andrewtweber:
Added a skeleton puppy

* Drops a puppy skull and bones when killed
* Puppy skull + 5 bones can be necromanced
* can only spawn in TX and Catacombs
* added 2 guaranteed to spawn in the bottom level of Catacombs

I need some help getting them to spawn. I've only come across one that randomly spawned in about 2 dozen attempts (using wizard mode, mostly in the Catacombs). Ideally I think they should spawn pretty frequently since they are unique to those 2 locations and fit with the location themes.


fejoa:
The PNG graphics diff works good :)

PR #294, opened 2017-11-26
Add new items II

red-kangaroo:
This PR attempts to balance new items added in last release, buffing those too weak, but decreasing frequency of many magic items, because they are spawning too often now. New items are also added, with full changelog as follows:

* Buffed chilling axes and scythes of terror, because they were triggering way too rarely.
* Buffed regeneration. It no longer requires you to be at maxHP for it to trigger.
* Item graphics from CLIVAN ported.
* Beast claws from alien mod ported, they are rather rare weapons using unarmed combat skill.
* Mangos and sausages added!
* Masks, a new headgear, added. Necromancers wear them.
* Taiahas ported from CLIVAN, but they don't explode when broken (they were unusable as weapons, which is stupid for a weapon) and are not normally generated - instead, they will be given as a quest reward in Mondedr.
* Scroll of body switch can be wished for.
* Cauldrons added, but not randomly generated. Can be used for special rooms (I have a vault with cauldron of healing liquid in mind ).
* Some potted plants and clocks added to the Cathedral, to make it more homey.
* Magic weapon price increased, per fejoa's suggestion. (Only of the new weapons, they had very low price.)
* Magic weapons and artifacts made more rare, because they were spawning way too often.
* New Fearless (Panic immunity) and Fasting status effects.
* Ring of bravery!
* Artifact armor "filthy tunic of martyr Bessica" grants disease immunity, fearless and fasting.
* New horn of healing and horn of plenty.
* New wand of webbing.


Main/Source/char.cpp
```diff
@@ -2363,7 +2383,7 @@ void character::GetPlayerCommand()
   {
     game::DrawEverything();
 
-    if(game::GetDangerFound())
+    if(game::GetDangerFound() && !StateIsActivated(FEARLESS))
```
andrewtweber: So `fearless` is intended to be a double-edged sword? You will never panic, but you also aren't warned about dangerous enemies? I like it :)

andrewtweber: Micro-optimization, but I think checking for the fearless state before `game::GetDangerFound` might save some unnecessary cycles. Same in other places

Main/Source/char.cpp
```diff
@@ -5016,6 +5032,34 @@ void character::PrintEndVampirismMessage() const
     ADD_MESSAGE("You recall your delight of the morning sunshine back in New Attnam.");
 }
 
+void character::PrintBeginFearlessMessage () const
+{
+  if(IsPlayer())
+    ADD_MESSAGE("You feel very brave.");
```
andrewtweber: I think this is a little understated, considering the player is now feeling a little reckless (ignoring highly dangerous enemies). I'll try to come up with a suggestion. Maybe something about `blood coursing through your veins`

Main/Source/gear.cpp
```diff
@@ -28,6 +28,13 @@ v2 meleeweapon::GetWieldedBitmapPos(int I) const
 { return SecondaryMaterial->GetVolume() ? item::GetWieldedBitmapPos(I) : v2(160, 128); }
 void meleeweapon::InitMaterials(const materialscript* M, const materialscript* S, truth CUP)
 { InitMaterials(M->Instantiate(), S->Instantiate(), CUP); }
+truth meleeweapon::IsRuneSword() const
+{
+  if(GetConfig() == RUNE_SWORD)
```
andrewtweber: Could be simplified to `return (GetConfig() == RUNE_SWORD);`

Script/char.dat
```diff
@@ -4817,13 +4822,14 @@ gibberling
   CanBeGenerated = true;
   Sex = UNDEFINED;
   CanUseEquipment = false;
-  CanTalk = false;
+  CanTalk = false; /* can, but no-one understands the gibberish */
```
andrewtweber: Would be funny to have it spew random "words" when spoken to

Script/define.dat
```diff
@@ -631,6 +640,15 @@
 #define BRAVERY 1
 #define FEAR 2
 #define CONFUSION 3
+#define HEALING 4
+#define PLENTY 5
+
+#define POTTED_CACTUS 1
+#define POTTED_PLANT 2
+#define DEAD_FISH 3
+#define BONE_FISH 4
```
andrewtweber: I think dead_fish should turn into bone_fish when it's eaten

andrewtweber:
There are still some of the newer items that are very underpriced. For example the blessed shield of the Phoenix sells for 45 and can be bought for 148 gold.

How do they get the filthy tunic? It seems a little overpowered. I would remove the Fearless state from it, and I think 10 AV is high for just a "tunic"

andrewtweber:
Just as a general note, I would be wary of adding too many special horns. Armor/weapons have slots and you're forced to choose even if you have many available, but you could theoretically end up with every horn and use them at will. But at least they have a "recharge" period

Main/Source/char.cpp
red-kangaroo: Yes, you become a bit too brave. :P

Do you mean only switching the order to '!StateIsActivated(FEARLESS) && game::GetDangerFound()', or moving the 'FEARLESS' check to 'GetDangerFound'?

Main/Source/char.cpp
red-kangaroo: Maybe we could go with understatement all the way? :D You feel a little bit more brave.

Or maybe You feel completely undefeatable!

Script/define.dat
red-kangaroo: Maybe you swallow the bones as well? :P

I'll try to do that, though. :) I like it.

Main/Source/char.cpp
andrewtweber: I meant switching the order. I don't think it's really important though. Optimization is just always on my mind.

red-kangaroo:
Hm, I'll check the artifacts and magical items once more, then.

The tunic needs a nerf, then. I would rather nerf the AV, though, because ommel bristler armor already has the spot for high-AV artifact armor, so the tunic should be more of a many-status-effects artifact armor.

Yeah, I'm wary of that but I find the way horns work interesting. I tried to balance them at least a bit by weight - horn of healing for example weights over 8000 g. Maybe I should make them unSOCMable, so that your horns will always be heavy? I could also decrease their possibility, as there are more types now.

andrewtweber:
I think nerfing the AV is sufficient, and also makes sense since it's just a tunic. Maybe 6 AV? That's high enough that with the additional bonuses + lack of dex/agi penalty, it is still an appealing choice compared to other special armors

Main/Source/char.cpp
fejoa: You shout: \\"I am invincible!\\"

(Monty python black knight joke)

Script/item.dat
```diff
@@ -5038,3 +5366,172 @@ ullrbone
...
+  {
+    DefaultSize = 30;
+    DefaultMainVolume = 150;
+    HasSecondaryMaterial = false;
+    BitmapPos = 96, 336;
+    FormModifier = 25;
+    StrengthModifier = 75;
+    Adjective = "fish";
+    NameSingular = "bones";
+    NamePlural = "bones";
```
andrewtweber: Minor issue, it will say things like "A fish bones is lying here"

Ideally "Some fish bones are lying here" or the name would have to be changed to something singular like "fish skeleton"

Script/char.dat
red-kangaroo: I'll write down a note to make gibberlings shout out random gibberish continuously, similarly to kamikaze dwarves' singing.

Main/Source/miscitem.cpp
```diff
@@ -3806,3 +3810,147 @@ col16 trinket::GetMaterialColorC(int) const
...
+        if(CanBeSeenByPlayer())
+          ADD_MESSAGE("Luckily, it's empty.");
+      }
+
+      RemoveFromSlot();
+      SendToHell();
+      return true;
+    }
+  }
+  else
```
fejoa: @red-kangaroo try removing this `else` statement to prevent the error in Travis and just leave `return false` where it is, outside the curly braces.

Main/Source/miscitem.cpp
```diff
@@ -3656,3 +3748,209 @@ alpha ullrbone::GetOutlineAlpha(int Frame) const
...
+    return false;
+}
+
+truth trinket::RaiseTheDead(character*)
+{
+  if(GetConfig() == BONE_FISH)
+  {
+    ADD_MESSAGE("%s suddenly comes back to life, but quickly air-drowns again.", CHAR_NAME(DEFINITE));
```
andrewtweber: I think "suffocates" would be better

Main/Source/miscitem.cpp
red-kangaroo: Travis is weird.

Main/Source/miscitem.cpp
red-kangaroo: Why not. :) "Air-drowns" sounds cool to me, but "suffocates" is probably more normal.

andrewtweber:
Looks good to merge to me 👍

andrewtweber:
@fejoa after this is merged how about a quick release? I think there have been several key (relatively minor but annoying) issues fixed since the last release so it'd be good to get a new one out
- siren blocking entrance to Cathedral
- overpowered goblin berserkers and zombies
- wishing for certain items

fejoa:
Yes definitely there will be a release before Christmas, just fitting development in when I can. I will try for a release by 00:00 GMT on the 23rd... wish me luck!
In the meantime this PR looks good to merge as well.

Issue #185, opened 2016-09-20
Undead based on strong monsters spawn too early


serin-delaunay:
In 0.50.5+, a lot of my runs have been ended in UT by zombies and skeletons based on powerful monsters. These have included an orc squad leader zombie, a skeleton of an elder dark mage (which was invisible for some reason), a zombie of a high-level mistress (can't remember which type), and countless veteran kamikaze dwarf zombies/skeletons. Some of these have spawned in UT1 before I've picked up any equipment, and most of them are powerful enough to stomp a typical character that reaches GC1.

I think these monsters spawn at low enough danger levels for the resulting imbalance to be considered a bug.

fejoa:
Undead configs are created here: https://github.com/Attnam/ivan/blob/master/Main/Source/char.cpp#L5904

Maybe we can change the way the base attributes are modified for each undead configuration?

andrewtweber:
Is it possible to only spawn undead versions of the living monsters that are spawning?

serin-delaunay:
Restricting undead spawns to those based on living monsters that are spawning might be overkill - there are a lot of monsters who shouldn't spawn for a given danger level, but whose undead versions are weak enough to be appropriate.
Still, unless undead are somehow bypassing the danger system, that might be the simplest fix.

AlexMooney:
The chameleons also have an annoying tendency to polymorph into high level ghosts, which is likely related to this.  Somehow the undead are not getting a very fair danger level evaluation.

andrewtweber:
I don't think the chameleons are necessarily related. They often polymorph into high level golems, etc. I think chameleons just polymorph into anything

andrewtweber:
I might be imagining it but this seems improved. I've still encountered zombies of what were once powerful monsters, but they're not as dangerous

red-kangaroo:
I have yet to encounter an out-of-depth undead in 0.50.9, too. Zombies of master necromancers or mages still appear, but they are not actually dangerous - without their magic, they have quite low stats. So it seems to be working right, for whatever reason.

andrewtweber:
Actually right after posting the above comment, I did encounter one. And I've still encountered a couple here and there. Hoping that fejoa's PR will fix this. For now I'm going to close this issue and we can reopen if it's still a problem

PR #295, opened 2017-11-27
DangerMap initializes zombies with all body parts

fejoa:
Causes the DangerMap to only consider zombie configurations that have all their body parts. This should have the effect that powerful zombies no longer appear "out of depth", but generate appropriately according to the IVAN danger system.

As a check: go into wizard mode and press '*', and check the consistency between naked danger, and equipped danger values, before and after this change.

Zombies will still generate with body parts missing, as normal.

Affects #185. Probably not a complete fix, but should go some way to alleviating the zombie kamikaze dwarf problem.

Can someone please check if it does what it says on the box?


andrewtweber:
Awesome! Looks good to me, the values are consistent and seem more appropriate

Issue #298, opened 2017-12-07
Debug log when game seg faults


andrewtweber:
The game seg faults here and there, but I don't have any useful information to report. It would be nice to add a backtrace and at least output that to the screen if not a log file, to make it easier for players to submit detailed bug reports.

Possible solution:

https://stackoverflow.com/a/77336/704803

Maybe this could be enabled with a flag, similar to wizard mode, so that casual players don't see it?

PR #300, opened 2017-12-15
Print stack trace if the game crashes on *nix

emlai:
or is quit in an abnormal way (excluding SIGKILL which cannot be caught).

Addresses #298, but only for *nix platforms, not Windows (because I had problems setting up the dependencies for my Window build). Not 100% sure if I'm trying to handle too many signals (didn't test them all), or if there are still some missing that we want to handle.

Tested on:
- [x] Linux
- [x] Mac

### DIFF
--- a/Main/Source/main.cpp
+++ b/Main/Source/main.cpp
@@ -17,6 +17,14 @@
 #include <sys/farptr.h>
 #endif
 
+#ifndef WIN32
+#include <csignal>
+#include <cstring>
+#include <cstdlib>
+#include <execinfo.h>
+#include <unistd.h>
+#endif
+
 #include "game.h"
 #include "database.h"
 #include "feio.h"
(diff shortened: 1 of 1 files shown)


andrewtweber:
Worked for me, but also on Mac

fejoa:
It compiles and runs on linux (Debian 8.2) only I can't remember a good way to get the game to seg fault.
Any hints?

andrewtweber:
I think one thing that works is to start a game, save & quit, edit the data files (or for example checkout red kangaroo's branch), make install the new data files, then try to load the existing game.

fejoa:
Ok I figured out a good way: set `EarthquakesAffectTunnels` to `false` in `ZOMBIE_LEVEL` in `GloomyCaves.dat`.
I tested on linux and the game just hangs when I pray to Silva on the zombie level.

emlai:
I tested with `*(int*)0 = 0;` added somewhere.

Yeah, unfortunately no way to detect hangs AFAIK.

fejoa:
Tested, and it works if I call IVAN from the linux terminal. Great stuff!
Thanks for the evil hint @emlai

PR #301, opened 2017-12-16
Make headless characters blind

red-kangaroo:
Previously, you could still see normally once you lost your head (and survived). Now zombies will now be blind without a head.


andrewtweber:
screen shot 2017-12-18 at 10 53 37

It's interesting that you can still at least see the path you've walked and items that you've passed. 

It's not total blindness like you would be without a light source, but I think that's probably fine since a blind person would be able to touch items and identify them. I wonder if it's possible to also "turn the lights out"?

red-kangaroo:
It might be possible, but I have no idea how. I thought this OK, as it makes no difference for monsters whether they are in the dark or can see only their own square, and for the player being unable to see anything beyond their own square is inconvenient enough. :)

Issue #299, opened 2017-12-14
Elianise not letting the player in the cathedral


badbeaver:
So, I just got to the cathedral of Attnam and she is just staying in that spot and not moving at all.
![ivan_2017-12-14_21-21-45](https://user-images.githubusercontent.com/34551546/34008069-2fd1540c-e115-11e7-829d-3e007141da4c.png)
![ivan_2017-12-14_21-12-59](https://user-images.githubusercontent.com/34551546/34008075-34b4727e-e115-11e7-818a-9623ace50442.png)

red-kangaroo:
There is a bug in siren AI, unfortunately, that prevents them from stopping trying to sing to adjacent creatures. Elianise wandered into the doorway and is now singing to the guards.

It was reported and fixed by vasiliy:

http://repo.or.cz/k8-i-v-a-n.git/commit/ae7c974ab2b4e6203f8558e10d2a29b472dea8e0

His fix should be rather easy to port to vanilla, IMHO.

PR #302, opened 2017-12-18
Fix sirens always trying to sing

red-kangaroo:
Based on k8IVAN, sirens now can fall back onto normal humanoid AI when their songs fail.

I also fixed their singing being "visible" from anywhere on the floor. Now they only can be heard.


red-kangaroo:
Fixes issue https://github.com/Attnam/ivan/issues/299

PR #308, opened 2017-12-26
Give shields AV

red-kangaroo:
Shields should give AV to all body parts, similarly how cloaks work. That way, you can either go dual-wield route and have more attacks/higher damage, or go shield route and have less damage, but more blocking (already present) and more AV (my change).

Even better, this change makes both light and heavy shields viable and different. Light shields are more accurate, thus better at blocking blows and negating the damage completely. Heavy shields are inaccurate and bad at blocking, but have much higher AV bonuses and will reduce damage taken.

Coincidentally, this also fixes a bug with Aegis which was already present in CLIVAN and I didn't notice it until now - it only gave its elemental resistances to the arm you wielded it with, not your whole body. Now it will be fixed. 

Discussion thread [here](https://attnam.com/topics/Shields-giving-AV). Based on chaostrom's idea, even though I forgot it was his idea. ;)


andrewtweber:
I would still like to see bashing added as a mechanic, to temporarily stun your enemy.

But this definitely makes shields more valuable. Looks good to me

PR #311, opened 2018-01-17
Update changelog

red-kangaroo:
I merged and updated NEWS file with all the changes since the new development, including changes in CVS. Now the players might get a better idea of what was changed. :)

I also suggest that from now on, we add a line (or more) into the NEWS file for each pull request, so that we can have a nicely arranged list of changes for each release. That would be much more neat than what we were doing so far. :D


andrewtweber:
I'm not a fan of removing the original Changelog, it's like a little piece of IVAN history :)

fejoa:
I agree we should preserve history, but I can see the need for consolidation. Maybe we can [automate changelog generation](https://github.com/skywinder/Github-Changelog-Generator) at each merge step? [We can preserve history by using the `--base` feature](https://github.com/skywinder/Github-Changelog-Generator#migrating-from-a-manual-changelog).

red-kangaroo:
@andrewtweber We should still update NEWS, because there was no changelog file since CVS. We could keep old Changelog file there as well, but my version is much easier to read. :D

andrewtweber:
Agreed. How about a folder called `historical` or something, which has the untouched Changelog, README, etc. That way they don't clutter up the main folder and it's clear that they're not updated?

red-kangaroo:
That's a good idea! There actually already is folder Doc/Obsolete, so we could move all files we deem redundant there.

fejoa:
Looks ok to me, @andrewtweber ?

PR #312, opened 2018-02-25
Imported the sound events amazing work from Ivan3D

AquariusPower:
hi! 

it is compiling and working on linux!
could someone test on windows and adapt related config files?

as SDL or SDL2 is an automatic configuration, my only doubt was the file "Main/CMakeLists.txt" that I had to put "SDL2_mixer" there to make it work. But as I was unable to compile it for the older SDL, I think it is not a big problem.

anyway, I am new to cmake config files, so I did what I could. If someone think it could be better done, it is probably true :)


AquariusPower:
uh... the travis builds failed? is that needed? what can I do to make it work?

andrewtweber:
If you click on the Travis CI link and then expand any of the builds it will show you details.

For example: https://travis-ci.org/Attnam/ivan/jobs/345858341

```
In file included from /home/travis/build/Attnam/ivan/audio/audio.cpp:36:
/home/travis/build/Attnam/ivan/audio/../Main/Include/message.h:58:10: fatal error: 
      'SDL_mixer.h' file not found
#include "SDL_mixer.h"`
```

andrewtweber:
Yeah check `.travis.yml` - you will see that SDL2 is being installed via apt/brew. You will need to add SDL mixer in there as well.

fejoa:
I made a `sound-effects` branch on the main IVAN project. We can use this to knock the cmake file and yaml files into shape

PR #319, opened 2018-03-01
TX balance patch

red-kangaroo:
This patch tries to address an issue TX has with game content and difficulty. Now that it lost the unfairly deadly golems that killed most characters, it's actually a bit easier than GC,  IMHO. Plus it lack certain content that makes GC more interesting. Thus:

* I buffed Xinroch's gravekeeper a bit. He should be a miniboss comparable to the first mystic dark frog in GC, not a simple renamed dark knight.

After running several tests against Elpuri and Oree, I found out that both forms of Xinroch are a bit lacking against the GC bosses.

* Ghostly Xinroch got full equipment, not just a sword - I mean, he has 400 HP, whereas Elpuri has 2000 HP, plus Xinroch is a humanoid...

* Resurrected Xinroch was better, but still got unarmed skill for when his sword is lost, because he could do nothing without it.

* Lost ruby sword now gives gas and disease immunity.

* Added hidden vaults to TX. GC has several, TX has none, which is not fun. They are different from the GC vaults, so enjoy the diversity. :)

* Added a small vault to UT and Attnam Catacombs, they should be rather hard to find and give you less awsome items than GC and TX vaults, but they add a bit fun to exploring those two dungeons.

All vault chests are locked, so you cannot simply get to the goods as soon as you get a pickaxe.

* Dwarven gas chambres can no longer be prematurely exploded by enner children.

* TX now has a small shop. Much smaller than Merka, but still a shop. :)


AquariusPower:
I applied the patch successfully, do you happen to have a test savegame? :)

Script/char.dat
```diff
@@ -1841,6 +1868,7 @@ ennerchild
   AttachedGod = SCABIES;
   BiteCapturesBodyPart = false;
   IsAbstract = true;
+  Inventory == Random { Category = RING|SCROLL|WAND; }
```
fejoa: Suggest ring or scroll only. Wand may blow up.

fejoa:
Great work, fingers crossed that nothing breaks, but it should be alright.

red-kangaroo:
I hope so, too. :D

PR #321, opened 2018-03-02
Outlined graphics and a few new sounds.

AquariusPower:
Imported the cool outlined graphics from Ivan3D.
Adapted the new existing graphics to be outlined too.
Also added a documentation on how to create the outline gfx properly.

Imported the cool eat.wav from Ivan3D-11, renamed to eatWell.wav.
Added some more new simple sound effects.


red-kangaroo:
Can we have a toggle in options to enable/disable those outlined graphics? Because it changes the look of the game quite a bit and does not look as smooth and natural as the non-outlined tiles. Players should be able to decide if they want a better readability through outlined tiles, or better-looking (YMMV) smooth graphics. :)

AquariusPower:
@red-kangaroo fair!
I added command line option: --outlinedgfx
I prefer the outlined because it is easier on my eyes, the non outlined gets confusing (like losing eye focus and mind concentration) after some time playing :>

fejoa:
Can the outlined graphics config option be set in IVAN's config menu, in addition to the snazzy command line arguments?

Does this change mean that two sets of graphics will need to be maintained? (I suppose it's not hard to draw things twice until a converter utility comes along).

AquariusPower:
I tried to put the graphics on the in-game config menu, but while reading the code I saw that the images array was quite limited and hardcoded, in a sense that adding the new file reference (instead of replacing the old one) would require changing many places, not 100% sure tho.

Also, the point on using the in-game config menu is to have changes applied without having to restart. That for the graphics would require a reload of that file. As I remember after it is loaded it is split into the actual items and beings etc, and that code could have to be isolated to be recalled cleanly to not create trouble, not sure tho.

So I ended up deciding on just replacing the reference thru the command line option, considering the lot more work required to let it be dynamically changed.

But I still think that, an in-game option could work, as long the user restart the engine. I am not sure tho, because of the moment the config file is loaded must be before the graphics files are loaded and split. Need to test it.

I will take a look if I can: let the outlined be changed in-game option just requiring a restart, do you think that requirement be ok?

**Graphics Maintenance:**
Anyway for now, I think whoever create/update new graphics, dont promptly need to do the outlined one. Just copy the default on the outlined file to not look like a missing graphics, and me or someone else later may create the outline. A very few graphics must never be outlined also, but these are exceptions, like the lanterns as I remember.

AquariusPower:
it took less time to code the in-game option than to speak about it, lol :)
I removed the command line option.
A game restart is still required.

PR #327, opened 2018-03-23
Enhanced scaling/stretching graphics using XBRZScale code.

AquariusPower:
here is the original code: https://github.com/atheros/xbrzscale
I transformed it into a library to be used with ivan.

For now, it only works with the LookZoom mode/option/key.

But I think it may be possible to use on the whole game, I am just not sure about performance then...
So, the xbrz library, for the whole game window, may require to be heavily modified to provide the maximum performance, for ex. I had to convert from 16bit color to 32bit to let xbrz work on it, and that may even be the key point on adapting it (let it work directly in 16bit), not sure tho...

here the results :

ivan 0 51_049

PS.: I removed the MAME exception clause for ivan, as you can read that they allow it there:
https://github.com/atheros/xbrzscale/blob/master/xbrz/config.h
https://github.com/atheros/xbrzscale/blob/master/xbrz/xbrz.cpp
https://github.com/atheros/xbrzscale/blob/master/xbrz/xbrz.h


fejoa:
It looks really nice!

AquariusPower:
uh... having some hard time to understand travis errors? how do I `make` igor and `mihail` (and what are they?)

fejoa:
I tested this and it works too. Friend, if you can make an option zoom the whole playfield with this feature to an arbitrary level, your tales will be sung in the halls of the devs for generations to come.

AquariusPower:
@fejoa xD

I just tried doing it, seems not so simple as I thought, here is some very bad initial test code at https://github.com/AquariusPower/ivan/tree/WideMonitorLayout (it draws over itself, unusable, just to see something happening...)

From that test, my first thought is that it may be better to use both stretchings:
- the simple and fast for effects (rain, snow, shining)
- and the xbrzscale for everything else. (may be shining, explosions and some other could be here too)

And, may be, the stretching should be made for every specific bliting and not after it all, because of the transparencies, but this is just a guess and I will have to make more tests.

fejoa:
Awesome! It's very promising, xbrzscale seemed quite fast, and it even animates the raindrops in New Attnam, along with Kenny's tail wagging. It makes the graphics look very amazing!
It reminds me of some of the scaling algorithm work done to spruce up the graphics in Star Control 2: http://sc2.sourceforge.net/screenshots.php

AquariusPower:
@fejoa I did some proper initial tests and it worked quite well, it was much simpler than I thought, more like what @Asmageddon said.

I think the speed is still good,  despite a bit slower.

There are many adjustments required to let it only work when the dungeon is actually visible, to not interfere with the main menu and all the dialogs like equipment and inventory etc.

This is scaled 2x with xbrzscale (yes I had to 1/2 w&h of the size of the visible dungeon area):
ivan 0 52_052

working here:
https://github.com/AquariusPower/ivan/tree/EnhancedScaling2ndPart

I am still not sure it will be the final thing, but it is already somewhat playable (need to fix the interference with dialogs).

Asmageddon:
Nice, this is looking really nice!

AquariusPower:
@Asmageddon 

that would only work (I guess) if each square was xbrz scaled independent, so we could determine what layer would receive xbrz. 

The workaround I found was to only xbrz around the player, so the remaining of the dungeon is fastblit scaled and looks like vanilla graphics. 

It speeds up a huge lot the processing and game gets practically as fast as vanilla, but you will still see the "out of view" with xbrz differently from w/o it. 
Sometimes it looks cool that difference, other times it is a bit annoying because that layer with xbrz looks quite different despite being the same thing near it w/o xbrz.

I guess, if we manage to draw the surroundings xbrz in a circle (instead of a square) it will blend better and be less contrasting.

But a real rework xBRZing each square, would be the real deal, one day :)

Asmageddon:
@AquariusPower What if you drew the map first, before applying the out-of-view overlay? Or used intermediate targets and layers?

As for updating the map, if I got your explanations right, the game only refreshes updated parts of the screen, right? Perhaps you could maintain unscaled graphics in one buffer, and upon being modified figure out which tiles need to be rescaled with xBRZ as well, otherwise just panning it?

But yeah, if the code is a bit spaghetti, that could be pretty painful to do :S

Either way, this is already pretty awesome, great job!

PR #341, opened 2018-04-17
(ready) Show items under and many other changes (another big patch)

AquariusPower:
EDIT: wherever there is a "dynamic" option, it is a in-between more complex/customized option than the others: not so animated neither so little animated, not so fast neither slow etc. Like the "thrown rotating items" that will rotate more initially/near the player.

(EDIT: ugh.. it is only commits from 83f4067 not all these shown there (that were already merged) it can be properly compared here: https://github.com/Attnam/ivan/compare/master...AquariusPower:ShowItemsUnder) (just squash it later if needed)

Here is what it does (above head or on corners V/H, all easily optional), mostly useful when you are "going" and stop above something and have to read the log or try to get things:

showitems2

**Other fixes/improvements/options (from commit log):**

Changes:
-    improved sfx regex for FallDown.
-    Pet inventory now uses xBRZ list item and silhouette.
-    When polymorphed, if not humanoid, will disable silhouette zoom.
-    During a polymorph question, will xBRZ list item and show silhouette.

Fixes:
-    DisplayStethoscopeInfo() fix to disable listItemScale and silhouette
-    Fixed list of selectables concerning hotkeys (limit is now the real 26 a-z recognized keys).
-    Fixed list of non-selectables page length for the entries count's user option (like the history log).
-    fixed a bug when changing dungeon scale, by updating the camera position just after loading the game, puts the player back on visible dungeon area, but also protected against player outside that area for ShowItemsUnder and xBRZAroundPlayer.
- Fixed xBRZ countours blending with transparency glitch.
- graphics.cpp: fixed access to SRegion vector indexes;
- StretchBlitXbrz(): allowed partial blit w/o crashing if remaining src image goes outside of dest image;

New options:
-  added look mode to alternate death reviewer 'l'. see full map only if wizard mode is compiled 'x'.
-  added user cfg option to show god extra info/about when praying.
-  alternative silhouette, with static tall and animated modes (the vanilla silhouette is still visible). If player is hopping or flying, related animations happen beyond the default breathing one.
-  rotating thrown weapons
-  show hits: from weapons and bodyparts (with new graphics for bite, kick and punch)

For many of these other things it would be better to create a video demonstration (instead of an animated gif) to show how all code flows look like. For now, just play it to see them.

Obs.: there may have other things I havent updated here yet, the commit log has it all.

PS.: if for any reason I stay too long w/o replying and it is working, feel free to squash merge it :)


fejoa:
I get a segfault if the scale is set to `x1`, and I select `UR, V` and then run over some bananas (ridiculous description to reproduce an error, I know)

AquariusPower:
confirmed:
```
./bin//ivan(_ZN6square24SendStrongNewDrawRequestEv+0xc)[0x7281f8]
./bin//ivan(_ZN4game26UpdateShowItemsAtPlayerPosEb+0x568)[0x7196c0]
./bin//ivan(_ZN4game20DrawEverythingNoBlitEb+0x63f)[0x718aa9]
./bin//ivan(_ZNK9character14ShowNewPosInfoEv+0xe7)[0x65a65f]
./bin//ivan(_ZN9character4MoveE2v2bb+0x53c)[0x64f306]
./bin//ivan(_ZN9character7TryMoveE2v2bb+0x900)[0x6505ec]
./bin//ivan(_ZN9character16GetPlayerCommandEv+0x20d)[0x656e65]
./bin//ivan(_ZN9character2BeEv+0x5ed)[0x64ec71]
./bin//ivan(_ZN4pool2BeEv+0x3b)[0x646e3f]
./bin//ivan(_ZN4game3RunEv+0x3da)[0x71800e]
./bin//ivan(main+0x455)[0x6ada23]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf0)[0x7f54a6a5c830]
./bin//ivan(_start+0x29)[0x642969]

```
I think it is because the drawing would happen outside DoubleBuffer, will take a look, thx!

AquariusPower:
mmm... the minimum xBRZ stretching is x2.

The dungeon w/o stretching is vanilla x1.

"Above head" using the minimum x2 stretch together with vanilla dungeon x1 looks messy as it fully hides many dungeon squares behind it.
This means dungeon x2 and above head would also look messy as would still fully hide one square.

To prevent the mess, I think it would be better to put constraints like:
Show items under should only be allowed if dungeon scale is minimum of x3, so the above head of minimum x2 would look correct (smaller than the square).

**But the "mess" is just my opinion, what do you think?**

Also, when using the corner, what actually happens is that the items are drawn overwriting dungeon squares. That's why they are not xBRZ when using the "around player xBRZ" mode, because the "items under, at corners" will be stretched together with the dungeon area using vanilla simple stretching code, what doesnt look cool (w/o xBRZ) but may be faster.

So, the other option would be to create a region to show things at corners, but they would be at least x2 for xBRZ to work. And then, they would NOT overwrite dungeon squares, but to keep the performance they may need to be limited to x2 xBRZ stretch. This means that, if I change it to use xBRZ while on corners, and the dungeon scale is ex.: x5, the items on the corners would still be x2, while now they are not xBRZ but look like x5 (same as dungeon scale).

This is still a blurred idea (on what would be better to be done considering performance) to me :)

AquariusPower:
let user decide!

above head working on x1 dungeon.

the real dungeon corners are used to draw the items for best integration/performance, but this means that to know the limits of the dungeon when it is all darkened you just have to cycle thru all corners... and that is a cheat :(
**Any idea how to workaround that? may be only allow above head on randomly generated dungeons?**

AquariusPower:
It is finally stabilizing, from now on (test/play time :)) I will probably just add fixes.

If you think something could be improved just say it and I will check what can be done, 
but for this PR, I will try to keep it simple so that it really stabilizes for the merging, unless it is something critical that I missed :)

It is interesting to merge this PR first: https://github.com/Attnam/ivan/pull/342
so users will be able to import savegames from the previous version: savegameVersion131,
But it is independent and may be merged later.

To test all changes from these latest PRs together just compile from this branch: https://github.com/AquariusPower/ivan/tree/THE_MERGED_BRANCHES_I_AM_PLAYING
I keep it updated always.

AquariusPower:
I think this one is ready.

I've been using this branch in conjunction with the wizard autoplay (from the other PR) a lot and detected minor glitches that are very hard to happen even in the fastest autoplay mode about hiteffects that are not cleaned when the camera moves and they havent finished playing (well... at least that is what it seems that is happening). I still dont know how to solve it but may happen only in wizard mode and with the autoplay enabled.

Main/Include/game.h
```diff
@@ -206,6 +206,13 @@ class game
   static cchar* Insult();
   static truth TruthQuestion(cfestring&, int = 0, int = 0);
   static void DrawEverything();
+  static void UpdateShowItemsAtPlayerPos(bool bAllowed);
+  static void UpdateAltSilhouette(bool bAllowed);
+  static v2 CalculateStretchedBufferCoordinatesFromDungeonSquarePos(v2 v2SqrPos);
+  static int IntemUnderCode(int iCycleValue);
```
fejoa: The code appears to be consistent, but you wrote `IntemUnderCode`. Did you mean `ItemUnderCode`?

fejoa:
Great new features in this one @AquariusPower. Can I merge?

Main/Include/game.h
AquariusPower: xD

AquariusPower:
The observations I did are just reference to be fixed (if we find how to) later.

Yes, it is good to merge.

PR #342, opened 2018-04-19
(ready) Savegame sort, dungeon progress basic info and old version importer

AquariusPower:
EDIT: added option to safely save the game: every file being saved, if overwriting, will first copy the existing one to a "*.bkp" new file. These temporary backups will remain there until the user successfully returns to the main menu, when all these temporary backups will be promptly removed. 
Why? if a crash or abort message happens for any reason, even if a blackout or OS freeze happens, these backups will remain there untouched. 
They can send them to us for more tests to play later when we fix the crash reason.
If the problem is unrelated to the game application (ex.: OS freeze or blackout), they can try to restore the backups manually.
Currently there is no support to auto-reuse the .bkp files, only to auto-delete them after a fully succesful save.
Another important thing about it is that it can also help on not requiring to give support to this work: https://github.com/Attnam/ivan/pull/344, as dup (or missing) player happens after a crash and trying to load from such savegames directly (instead of the safe backups).

DONE: EDIT: WIN32 and DJGPP code still needs test/confirmation that they are working (despite I believe they are working)

EDIT: the old savegame (v131 up) importer (optinal feature) test is possible when merging with https://github.com/Attnam/ivan/pull/345

DONE: EDIT: re-used WIN32 and DJGPP code (to fill the files list), just need confirmation they are working too.

**EDIT: if the sort option is disabled, it will fallback to the old code for safety (therefore it was not modified), so it is granted to work as it always did in the past.**

It looks like this with the reversed time sort (therefore the last savegames show on top) and dungeon ids as  a basic progress info (how far you went).
selection_084

The problem is... I could only test it at linux (therefore I only coded it for UNIX).
I guess WIN32 and DJGPP can re-use most of what I coded and just require a few changes, because what was already there had a few changes between each other.
So, I need help on providing support for WIN32 and DJGPP.
Anyone interested on it that has write access on this branch may help, if not, just drop a pull request, but I still can only test on linux...

EDIT:
I can try to blindly code later, but I will still not be able to test it.
btw, these are the options:
```
+  case 0: Entry << "no sort";break;
+  case 1: Entry << "alpha";break;
+  case 2: Entry << "alpha+ids";break;
+  case 3: Entry << "DtTm+ids";break;
+  case 4: Entry << "revDtTm+ids";break;
```

Obs.: if for any reason I stay too long w/o replying and it is working, feel free to squash merge it :)


AquariusPower:
mmm... I use this gcc that works:
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609

here https://stackoverflow.com/questions/14136833/stdput-time-implementation-status-in-gcc says std::put_time should be implemented in gcc 5, and at the failed linux compilations for travisci I read:
The CXX compiler identification is Clang 5.0.0 **(this should be working right?)**
The CXX compiler identification is GNU 4.8.4 **(this one is really incompatible, can it be upgraded?)**

AquariusPower:
portable file modification time tip from: https://stackoverflow.com/a/40504396

fejoa:
This is good so far, just a few questions and some stability feedback:

Does this only open saves that are savefile version 131 or higher? (i.e. ivan version number 0.52 or higher?)

On windows, if I choose `no sort`, then I can load savefiles from v129 (that's around ivan version 0.50), which crashes the game. No warnings show up that this might be a bad idea. Versions that are savefile v131 load without failure, which is great. I like the way the incompatible saves are grayed out in dark grey.

What do the little ID numbers mean? Would it be better to describe where the character is? Say "Gloomy caves level 3", "Underwater tunnel level 2", that sort of thing?

Is there any extra development maintenance needed when adding new monsters/items to ensure savefile compatibility?

AquariusPower:
> Does this only open saves that are savefile version 131 or higher? (i.e. ivan version number 0.52 or higher?)

yes, but future version may fail depending on how we code it:

> Is there any extra development maintenance needed when adding new monsters/items to ensure savefile compatibility?

The whole feature depends on having NEW savegame data stored AFTER everything else that is already there. 
So, if the developer wants to keep the good reading or clarity or good organization of what will be saved, and put new things in-between (and not appended), the importing will fail :(.

Surely appending may be done this way in case the new things are from many classes, we would have to create specific methods like this one on every class with new save data:
`SaveDataForV133()`
and so on, and create some kind of validator, to grant all new stuff is stored sequentially appended on the savegame file.
I think a better idea may come out later...

Therefore it is experimental, it worked now but may not work later w/o improving the binary saving flow.

Obs.: a overall better way would be to have tokens identifying begin and end of sectors, use a loop to detect each and delegate the loading, then we would be able to move all binary data around the savegame file freely, but that would require a big effort to prepare all the savegame code to do that I guess.

> On windows, if I choose no sort, then I can load savefiles from v129 (that's around ivan version 0.50), which crashes the game. No warnings show up that this might be a bad idea. Versions that are savefile v131 load without failure, which is great. I like the way the incompatible saves are grayed out in dark grey.

I dont know what to put there. The available space to write a good explanation message is too short. So instead of "no sort" do you think this would be better? "no sort (old saves will crash)"

We could keep just the new code with sorting, but I think that is not safe, mainly because user may type weird characters for the player's name like (non a-zA-Z0-9), and I didnt protect the code against it (mmm... well the old code isnt protected either but I think it is less prone to fail, not sure tho).

The old code is still there. The "no sort" option is just the old code! as a fall back in case anything weird happens with the new sorting code, preventing the user from loading the savegame, he/she just needs to disable the sorting!

>What do the little ID numbers mean? Would it be better to describe where the character is? Say "Gloomy caves level 3", "Underwater tunnel level 2", that sort of thing?

mmm... good! may be the short names for it would work well too like:
(GC:1,2,3;UT:1,2:TX:1,2,3,4)
the problem is the available space, may be this would look clenaer:
(GC 1 2 3, UT 1 2, TX 1 2 3 4)
what u think?
The full names wont fit I guess. Well, I have like 2 full pages of savegames, because of all the tests, but  I think most players have at most 2 saves, so yes the full names could be a good alternative. May be I put that in the options: full or short dungeon names?

fejoa:
Great, I understand more now, thank you for the explanations.

I think all we need to do then is to leave a comment in the code that new savefile items are to be appended only, and give a short example of what we mean.

Maybe we just change the colour of the old saves in the unsorted case, indicating that they may fail. 

Let's implement your clean version of short names. Is it possible to highlight in yellow the name and number of the current level that character is on? Or is that info impossible to determine without loading the whole file?

I like to grind characters through underwater tunnel and save in Attnam so I can grind them through GC later. So in this way I get long lists of characters as well.

AquariusPower:
>I think all we need to do then is to leave a comment in the code that new savefile items are to be appended only, and give a short example of what we mean.

yes, the biggest problem is when the change happens in a new file that has no such comments, the developer may miss doing it.
The best way I guess is to actually try to import an old save. As soon it fails, the developer will know/remember the new stuff should be appended.

But let me add this:
ClassA -> Save()
|   |_calls ClassB->Save()
|_continues saving ClassA stuff
|__calls ClassC->Save()

now ClassB has new stuff.
it cant be appended at the end of the ClassB->Save() method
it must come like:
ClassA -> Save()
|   |_calls ClassB->Save()
|_continues saving ClassA stuff
|__calls ClassC->Save()
|_ ClassB()->SaveV133NewData()

I havent tested all this, but I think it is what we will need to do to keep importing old saves compatibility.
(I think the idea on how to do it is getting clearer :))

>Maybe we just change the colour of the old saves in the unsorted case, indicating that they may fail.

red!

>Let's implement your clean version of short names. Is it possible to highlight in yellow the name and number of the current level that character is on? Or is that info impossible to determine without loading the whole file?

I tried to find a way to colorize only individual words on the messages log once, but failed.
I still dont know if there is a way for that, I think it is for the whole line only, the color option.
I would like to patch, if possible the code that writes the text, that would be great to highlight when we get hit or something bad happens, with red, yellow etc. A good way to filter all the text in the message log!

For now, where we are could be indicated like: GC 1 [2] 3, or GC 1 2 <3> or {1}  :) 

>I like to grind characters through underwater tunnel and save in Attnam so I can grind them through GC later. So in this way I get long lists of characters as well.

may be we all do that I guess? I always reach there and say: "mmm... time to think what to do now" xD

AquariusPower:
there is a big problem..

to show the dungeon/level description or shortDescription, 
we would have to pre-load each savegame file, collect all dungeons, initialize them enough to collect this information, because dungeon data is dynamic (the dungeon scripts may change w/o changing the code as I understood).

To merely collect the current dungeon location I had to skip the gamescript that contains all basic dungeon data.

I think, that is possible to be done, 
would be cool,
but may be overly complex...

The mere IDs are quite meaningless I know.

May be an issue to show full or short dungeons name and level progress based on reading binary data dynamically could be created.

But much better would be to create a simple progress description when the game is saved.
It could even be a new simple text filename, user readable, like: PlayerName.progress
That would be much better than what I did already I guess, because there we could put detailed but still just basic information about the progress, like full dungeon names and current level in roman :)

AquariusPower:
mmm... to change the color of invalid saves on the "no sort" mode, that actually is the old  vanilla code, I would have to reimplement the inital reading of binary data from savegame files there too...

It is much better to modify it to "no sort (old saves will crash)" and explain in some doc that it is the vanilla code being used to list the savegame files as a fallback in case something do not work well with the new fancy sorting code. 

Another option would be to modify the default to be the last option (reversed sort by dateTime) so the latest savegame is shown on top, and the user just have to hit Enter to continue playing! what about it? and it will have the darkened/unselectable old saves!

AquariusPower:
bad news :( 
(well.. not that bad..)

about the old savegame importer...

I was thinking, making ALL new stuff to be loaded sequentially from the file may make the code unbearably complicated/messy/confuse to maintain... and that is a bad thing...

Also, to create (string) tokens ex.: "Character:ID=123:MemorizedEquipment" to let every sector be placed anywhere and loaded from anywhere, may become overly complicated too... 
Lets say we increase the max equipment slots from 13 to 15 (adding wings for piercings for example), how that will be coded?
We would have to load 13 slots, then add separate code to load only 2 slots! and that is a huge mess... therefore a baaaad thing...

So, the best shot for now is: For every new savegame version, we create an specific importer to the PREVIOUS one ONLY! ex.: from v131 to v132 (added equipment memory), the importer will skip loading the equipment memory ONLY from v131 save files.

When increasing to save v133. We DROP the code importer for v131, so it will REQUIRE "equipment memory data" to be present on the save file.
And a new importer for v132 ONLY may be made available when the current save version is v133.
**So if things got too coplex to deal with, _such importer will not be available_, it will all depend on the current developer opinion about it.** (unless someone else helps on creating the importer)

I think that is the best way to keep it simple enough to lower the maintenance effort, and make it easier on all of us :)

AquariusPower:
@fejoa 

I am thinking on dropping the old savegame listing code.

The reason I kept it there was in case the new code fail and the user ends with no choices.

But, as you said, game will crash/abort if trying to load an old save with the old code, therefore the new code is not fancy improvements, but also to quality and stability.

So, in this case, the new save list would not be optional, but would be a replacement code, as it also works as a bug fix.

What you think?

fejoa:
@AquariusPower yes I think it is appropriate to replace the old savegame list code. I noticed a lot of your contributions have been configurable, but there also comes a time when these options are better than the default and should be the new default. Players will also notice new improvements more easily if they don't need to go through a configuration menu to discover them.  Please make the changes as you see fit : )

AquariusPower:
@fejoa  cool ok, I reworked it a bit to be more readable and made available what we may most use I guess: newest or alphanumeric, optionally with progress.

As I said, the progress can be improved later. These dungeon numeric IDs can become the shortened dungeon name, but a new PlayerName.progress or .log or .prg file will be required to be created during the save time! 
and we can add a lot of info there if we want, even the gameplay messages history log, or a shortened story removing all battle logs (having only travels, prayings, diseases, and the game story events like talking to Petrus or the necromancer etc).

Btw, all these naming, options text and help text that I created in all my PRs, feel free to modify to make it clearer to end user! I throw in the 1st naming that comes to my mind (like dungeon IDs lol), and these things may be confuse if the user is not a coder.

fejoa:
> As I said, the progress can be improved later. These dungeon numeric IDs can become the shortened dungeon name, but a new PlayerName.progress or .log or .prg file will be required to be created during the save time!
and we can add a lot of info there if we want, even the gameplay messages history log, or a shortened story removing all battle logs (having only travels, prayings, diseases, and the game story events like talking to Petrus or the necromancer etc).

Yep this sounds like a good idea. I agree, probably a separate feature to this PR that can be added later?

> Btw, all these naming, options text and help text that I created in all my PRs, feel free to modify to make it clearer to end user! I throw in the 1st naming that comes to my mind (like dungeon IDs lol), and these things may be confuse if the user is not a coder.

The descriptors / tool-tips you have chose for the configuration menu seem good enough to me. 

Ready to merge?

AquariusPower:
sure, I think it will not be complicated to add a simple text file with some identifiers like:
```
progress=UT: I @II III, XT: I II II IV V #despite I like it, not sure if this is that cool, may be just numeric?

eventsHistoryLog #all appended lines below
...
```
I am dropping the above idea here just as reference for later.

yes, ready to merge, I have found no other problems til now.

AquariusPower:
The savegames file listing is already usable (could be better later, but already works), so this PR can be merged only for that.

But the safe savegame protection is currently useless as it creates a backup before saving and does nothing with that backup to help the user...

The most crashes I have been getting latelly is when entering that caves with bears and lions on the 1st dungeon. The crash is random and happens when that level is created. This kind of crash seems to trash the savegame sometimes.

So, rethinking the safe savegame mode, I think it should save on new .SomeThing.tmp files, and only if the game does not crash, that file will be renamed to the real file (SomeThing: .sav .40 .41 .wm etc).

So if you would like to merge this PR, it is **ok** for the **file listing**, as will also already help users to prevent crash when trying to load old saves too.

But I will try to **rework the safe savegame** to make it useful, and also try to **link the AutoSave with the correct player name**, and if possible allowing more than one AutoSave.

PS.: I think we are not limited anymore to the old windows file naming restrictions "12345678.123" ? so windows now can deal with filenames like "PlayerName.AutoSave.40" etc. Because that flexibility will be required to implement it in the less complicated way possible (as the way I imagine it of course :)).

AquariusPower:
It is all good now!

If there is AutoSave (per player name now), it will be auto chosen and restored.

If there is not AutoSave but there is .bkp files, the user can chose to restore the backup.
If not, the game may crash with a message indicading the user should try to restore a backup.

I got a good test case for that, if you need I can upload it.

I put this info in the source code:
```
          /**
           * Backups are a fallback mainly when a crash happens like this:
           *
           * - USER action: you try to enter a new dungeon (so it will be created now)
           *
           * - game action: saves a backup of the current dungeon last save at ".bkp"
           * - game action: saves the current dungeon w/o the player on it at ".tmp" (so if this saving crashes, the corrupted .tmp will just be ignored)
           * - game action: copies the sane ".tmp" to the final filename (and deletes the ".tmp" just after)
           *
           * - game bug: the new dungeon creation fails and crashes.
           *
           * This means: the only file remaining with player data (a player char at dungeon pos) is the .bkp one!!!
           */

```

when it randomly happens, when going to gloomy caves the 1st time after playing a lot even at XT, I get this backtrace (about light emmiters, and this is about the 3rd time) I still dont know how to fix it!

```
  what():  std::bad_alloc

./bin//ivan(_ZN6dbgmsg20getCurrentStackTraceEbRi+0xaa)[0x898c7e]
./bin//ivan(_ZN6dbgmsg22getCurrentStackTraceSSB5cxx11Ebb+0x3f)[0x898d67]
./bin//ivan(_ZN6dbgmsg8SigHndlrEi+0x18e)[0x89940a]
/lib/x86_64-linux-gnu/libc.so.6(+0x354b0)[0x7f29ae8bf4b0]
/lib/x86_64-linux-gnu/libc.so.6(gsignal+0x38)[0x7f29ae8bf428]
/lib/x86_64-linux-gnu/libc.so.6(abort+0x16a)[0x7f29ae8c102a]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(_ZN9__gnu_cxx27__verbose_terminate_handlerEv+0x16d)[0x7f29af21184d]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x8d6b6)[0x7f29af20f6b6]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x8d701)[0x7f29af20f701]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x8d919)[0x7f29af20f919]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x8debc)[0x7f29af20febc]
./bin//ivan(_ZN9__gnu_cxx13new_allocatorI7emitterE8allocateEmPKv+0x40)[0x7a3070]
./bin//ivan(_ZNSt16allocator_traitsISaI7emitterEE8allocateERS1_m+0x28)[0x7a18e3]
./bin//ivan(_ZNSt12_Vector_baseI7emitterSaIS0_EE11_M_allocateEm+0x2a)[0x79e9ee]
./bin//ivan(_ZNSt6vectorI7emitterSaIS0_EE14_M_fill_insertEN9__gnu_cxx17__normal_iteratorIPS0_S2_EEmRKS0_+0x2ab)[0x7a0df5]
./bin//ivan(_ZNSt6vectorI7emitterSaIS0_EE6insertEN9__gnu_cxx17__normal_iteratorIPKS0_S2_EEmRS5_+0x88)[0x79e2cc]
./bin//ivan(_ZNSt6vectorI7emitterSaIS0_EE6resizeEmRKS0_+0x91)[0x79b0f9]
./bin//ivan(_ZrsI7emitterER9inputfileS2_RSt6vectorIT_SaIS4_EE+0x51)[0x797c9d]
./bin//ivan(_ZN7lsquare4LoadER9inputfile+0xbe)[0x77f0b8]
./bin//ivan(_ZN5level4LoadER9inputfile+0x24e)[0x7732f8]
./bin//ivan(_ZrsR9inputfileRP5level+0x45)[0x777cfd]
./bin//ivan(_ZN7dungeon9LoadLevelER9inputfilei+0x37)[0x8047bf]
./bin//ivan(_ZN4game17PrepareRandomBoneEi+0x258)[0x7476fa]
./bin//ivan(_ZN4game9EnterAreaERSt6vectorIP9characterSaIS2_EEii+0x55)[0x744247]
./bin//ivan(_ZN4game9TryTravelEiiibb+0x6f)[0x743f91]
./bin//ivan(_ZNK6stairs5EnterEb+0x433)[0x78e16d]
./bin//ivan(_ZN13commandsystem6GoDownEP9character+0xaa)[0x7638e2]
./bin//ivan(_ZN9character16GetPlayerCommandEv+0x4c0)[0x66c062]
./bin//ivan(_ZN9character2BeEv+0x5ed)[0x65bbab]
./bin//ivan(_ZN4pool2BeEv+0x3b)[0x65345b]
./bin//ivan(_ZN4game3RunEv+0x3da)[0x733802]
./bin//ivan(main+0x475)[0x6c851e]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf0)[0x7f29ae8aa830]
```

So, having the .bkp file, it let us try to enter the new dungeon again, and the crash may now happen, allowing continue playing it!

marmistrz:
Have you thought about using this image: https://docs.docker.com/samples/library/gcc/ ?

PR #344, opened 2018-04-21
(ready) BugFix/Workaround duplicated player after crash on loading autosave (there is a user option to enable this experimental code)

AquariusPower:
A new developer/advanced option allowing recovering the savegame using the old player instance (the one left behind on the level), or the new player instance (your latests items and stats, nutrition etc), or just disabled and the game will continue crashing.

The code is isolated in a new class for easy removal one day when this obnoxious bug never happens again.

here how it looks when the bug is detected, the user can chose when to apply the fix:
selection_089
selection_088
not allowing will abort the game with a friendly message.

if the option is disabled, the game will just crash as vanilla.

here a test case, you have to load the AUTOSAVE one:
[Save.CURRENT_TEST_CASE.zip](https://github.com/Attnam/ivan/files/1934091/Save.CURRENT_TEST_CASE.zip)

this is completely experimental. worked here with this only test case I got til now...

it basically moves the trashed player (old or new) to the monster team and removes all his items.
also, it ensures there is no duplicated items (and bodyparts) thru the whole level squares and NPCs.
you will see the trashed player with items initially but soon after all his items will vanish, so granted (as much I could) no duplicated items.

I didnt test anything else then killing it and ignoring the corpse, neither long play after doing it.

Obs.: if for any reason I stay too long w/o replying and it is working, feel free to squash merge it :)


AquariusPower:
I am dropping it in favor of:
https://github.com/Attnam/ivan/pull/342#issuecomment-390837685

Despite it was very helpful to learn how the internal database is made consistent.
May be this code can be isolated and moved to the Obsolete folder as useful reference, retrieval one day if needed, not sure tho, I will keep that branch anyway (may prepend "obsolete" to it).

AquariusPower:
from this report: https://github.com/Attnam/ivan/issues/445
I am reopening this PR to let travis-ci compile it and for further tests.
I isolated the workaround code to not complicate the already bloated game.cpp file, and be easily obsoleted one day.
It will just sit there until someone (desperately :)) needs it, as is an advanced option that needs to be enabled.

PR #345, opened 2018-04-23
Improved humanoid pet equipment management (incomplete)

AquariusPower:
**How it works:**

Basically this will let your humanoid pet to:
- whenever polymorphing to another humanoid, auto re-equip everything in correct places automatically.

Now, every item that cannot be equipped will be moved to the pet's inventory, and so can be easily equipped if it turns into a new humanoid polymorphed form.

Also, for now, **the equipped memory is still not stored on the savegame**, I think this will require a new savegame version making it incompatible with old saves, still dont know how to do that, have to research.

**considerations:**

I think this makes humanoid pets useful again, as it was quite a waste of time having to re-equip all its items, and I just ended giving up on trying and letting it die..

From what I read on the code, it was just a missing feature, that was only implemented to the player, therefore not saved on the player character data but directly on the savegame file. So I understand it is not a feature as it is now in vanilla.

In the other hand, to let all NPCs use this feature may change the gameplay too much, because when polymorphing all equipped items would be dropped if the new form is not humanoid (or cant use the item), what certainly was a tactical movement. So I limited this capability only to the pet for now (this may be an user option later and will make the game much harder I guess).


fejoa:
It works on windows. Ready to merge?

fejoa:
I tested with the slave, who I poly'ed into a ghost and dropped all equipment. Then I waited until the slave returned and when he walked over his stash, he equipped everything I gave him except his pick-axe which I equipped as his weapon. I re-equipped this manually. 
Later I poly'ed him again and he turned into an orc general and he kept all his stuff. It works really well!

One thing is that it increments the savefile version. Does merging this affect your other savefile improvement branch?

AquariusPower:
>I tested with the slave, who I poly'ed into a ghost and dropped all equipment. Then I waited until the slave returned and when he walked over his stash, he equipped everything I gave him except his pick-axe which I equipped as his weapon. I re-equipped this manually.

So you had minimal trouble, what is a issue. 
Despite I think the ghost should not drop the stuff, not sure tho..

I will try some more polymorphings to see if I missed something.

>One thing is that it increments the savefile version. Does merging this affect your other savefile improvement branch?

if we  merge this first, v131 saves wont be imported.
but as soon the other PR is merged, there is commented code for this PR to allow importing v131 (by ignoring loading v132 new data from v131 file (as it is not there)), so I think the order will not be a problem.

Issue #347, opened 2018-05-04
parasitized damage crash fix


AquariusPower:
Fixed this bug about parasitized damage:

```
Segmentation fault
./bin//ivan(_ZN9character12HealHitPointEv+0x2b)[0x676f79]
./bin//ivan(_ZN8bodypart13ReceiveDamageEP9characteriii+0x13e)[0x800566]
./bin//ivan(_ZN9character21ReceiveBodyPartDamageEPS_iiiibbbb+0x3ec)[0x66434e]
./bin//ivan(_ZN8humanoid13ReceiveDamageEP9characteriiiibbbb+0x4ff)[0x6f10f7]
./bin//ivan(_ZN9character18ParasitizedHandlerEv+0xcb)[0x676b35]

```
the code tracking:
```
void character::ParasitizedHandler()
... // the damager is the 1st param,value: 0
    ReceiveDamage(0, 1, DRAIN, TORSO, 8, false, false, false, false); // Use DRAIN here so that resistances do not apply.
...
  if(BodyPart->ReceiveDamage(Damager, Damage, Type, Direction)
...
    if(Damager!=NULL && (Type & DRAIN) && IsAlive()) //HERE IS THE FIX, Damager!=NULL
      for(int c = 0; c < Damage; ++c)
        Damager->HealHitPoint(); //HERE was the crash

```
This simple fix can be a separate pull request from this one where I fixed it:
https://github.com/Attnam/ivan/pull/341
commit: https://github.com/Attnam/ivan/pull/341/commits/6832723e5a078151f54b3108b14c922de7ec38dc

Seems ok right?

If you would you like I create another PR just for it, just say it :)

fejoa:
:+1:  Yes that would be great

PR #348, opened 2018-05-05
Fix parasite damage crash

AquariusPower:
https://github.com/Attnam/ivan/issues/347
EDIT: mmm... it is only one commit...


fejoa:
Presumably this works

AquariusPower:
xD
I forgot to say how to test: 
as I remember u just have to do not pay attention and eat a zombie by mistake, complain a bit and ignore the fact, until you move enough around to reach Attnam town when you believe all will get better and suddenly *CRASH*... but not anymore :)

PR #349, opened 2018-05-07
(ready) New monsters

red-kangaroo:
Adds several new monsters and uniques to spice up the game.

Also fixes spellcasting of Anmah and gives each archangel their own personalized equipment rather than generic equipment for alignment.


AquariusPower:
I couldnt recreate the error message, everytime it compiles successfully here.

I even tried adding these CXXFLAGS to no avail:
```
-Werror -Wno-format-security -Wno-undefined-var-template -Wno-unknown-warning-option # from mxe/IvanWin/build.sh
-Werror=aggressive-loop-optimizations # from travisci log at github (despite I think it is just -Werror
```

but, I think this could solve the problem at char.cpp:
```
truth character::CheckAIZapOpportunity()
...
  // 012
  // 3 4
  // 567
  int CandidateDirections[8] = {0, 0, 0, 0, 0, 0, 0, 0};
```
I dont know if I can trigger travisci to surely test it tho?

red-kangaroo:
Damn, every time and still it compiles normally. I guess I'll try your fix and see what Travis has to say on that, thanks.

fejoa:
Just a heads-up: the code for AI zapping/throwing is inefficient and could slow the game noticeably, so it will definitely need some improvement. I think it might have something to do with the radius which the NPC searches. Also the AI needs to check for walls, because I remember observing axe throwing dwarves in CLIVAN throwing objects at the wall in the direction of the player, when the player character was behind a wall...

AquariusPower:
I think I finally understood this error message:
"error: iteration 7u invokes undefined behavior"

ok, the array is still configured with 7 positions: CandidateDirections[7]
so "defined behaviors" (valid ones) are indexes 0 to 6.

When the loop `for(uint dir = 0; dir < 8; ++dir)` reaches index 7, 
this cool compiler (that I still cant manage to enable/configure here) will say that index is an "undefined behavior", therefore 7 (the "u" may mean this uint @fejoa just coded? that hinted me what the error message meant!)

I am quite sure changing to 
int CandidateDirections[**8**] = {0, 0, 0, 0, 0, 0, 0, 0};
will fix it now!

PS.: how that compiler for ivanWin is configured? that stuff is really nice! this option: "-Werror=aggressive-loop-optimizations" will not make it work here...

red-kangaroo:
How inefficient is the code? It does not seem worse than code for dark mages looking for a target, but I may be pretty wrong.

I think SquareUnderCanBeSeenBy should stop zapping at walls, because it does not permit any form of seeing through walls (infra or ESP).

fejoa:
Well the axe dwarves threw stuff so that may just have been the flying axe animation causing the lag, especially if something got thrown a long way. Zapping probably worked a bit faster as there is less to animate. 

Awesome work fixing the wall thing so simply by the way, I hadn't thought of SquareUnderCanBeSeenBy. There's plenty the old devs left behind to help us : )

red-kangaroo:
> Well the axe dwarves threw stuff so that may just have been the flying axe animation causing the lag, especially if something got thrown a long way. Zapping probably worked a bit faster as there is less to animate.

I hope so, but I'll try to check if anything is slower with the zapping code anyway.

Main/Source/item.cpp
```diff
@@ -1918,11 +1918,11 @@ void itemlock::PostConstruct()
 
 truth itemlock::TryKey(item* Key, character* Applier)
 {
-  if(GetVirtualConfig() & BROKEN_LOCK)
```
fejoa: May I suggest this #352 as a fix for this instead?

red-kangaroo: Nice, I'll revert my change.

AquariusPower:
I am not sure if I got a valid issue or not. 
I think it may be just a developer (minor) issue, as end users will probably always use the latest game version always.

I created a new char, and played a bit, 1st dungeon.
I could load it with or w/o this new monsters patch.

I created another char, and played a lot (using wizard autoplay).
I could only load it using this patch, crashing otherwise with:
```
	./bin//ivan(_ZNK18characterprototype12SpawnAndLoadER9inputfile+0x8a) [0x671e80]
	./bin//ivan(_ZrsI9characterER9inputfileS2_RPT_+0x5c) [0x6ac089]
	./bin//ivan(_ZN6square4LoadER9inputfile+0x37) [0x7ff581]
	./bin//ivan(_ZN7lsquare4LoadER9inputfile+0x69) [0x778b13]
	./bin//ivan(_ZN5level4LoadER9inputfile+0x24e) [0x76cda8]
	./bin//ivan(_ZrsR9inputfileRP5level+0x45) [0x7717ad]
	./bin//ivan(_ZN7dungeon9LoadLevelER9inputfilei+0x37) [0x7fe29b]
	./bin//ivan(_ZN7dungeon9LoadLevelERK8festringi+0xed) [0x7fdb41]
	./bin//ivan(_ZN4game4LoadERK8festring+0x824) [0x738d46]
	./bin//ivan(_ZN4game4InitERK8festring+0x207) [0x72f8c5]
```

I think this crash means it tried to load a char prototype that does not exist w/o this patch.

As I understand, all old savegames will be promptly compatible with this patch applied, and will be still compatible w/o this patch as soon no new monsters have spawned.

Despite there were no changes in the file `.sav` structure, **I think the savegame version should be increased**. 
And the "old savegame" importer (sector skipper) code will not be required as it is already compatible!

Why all that? if we use an older game version, that have the savegame sorter patch, it will block games created with this patch applied, preventing a crash in case there is any new monster spawned on the savegame dungeon files (btw, these dungeon file have no version as I understood).

_Unless I am wrong, and unless this is unnecessary :)_

**This will be unnecessary** if the next ivan release comes both with this patch and the savegame sorter patch, as the savegame version will be v132 for all these patches together.

If I am not wrong, these considerations should be added to some documentation, to tip devs on increasing the save version even if the `.sav` is not modified, but the dungeon files be modified in an incompatible way with old game versions (loaders).

fejoa:
> This will be unnecessary if the next ivan release comes both with this patch and the savegame sorter patch, as the savegame version will be v132 for all these patches together.

Just a general comment, we will probably add the v053 tag to trigger the next windows build sometime in June 2018. I think this PR, and all the other stuff will be merged by then. If the savefile version needs to be incremented for stability reasons, then that is okay.

AquariusPower:
I always thought like going thru the first 3 UT dungeon levels was getting boring cuz of too easy/predictable in a way. I think it was made like that to get easy on new comers. Now it feels they are much harder now (actually just because of golems (at least) like ash or pottery clay (I think it was that)), and... I really prefer it as it is now!

red-kangaroo:
@AquariusPower The golems are most likely because of #360 It added new low-tier golem materials, so they would generate more often and in more variety. Each type of monster rolls its frequency separate to see whether it should spawn, so more golem types with same frequency as before means more golems spawned. I probably should decrease the spawn rate of golems a bit to make up for their increased variety.

But I do agree with you that UT often feels easy. :D I have no idea if we made the early game too soft with our changes of if we all get hang of at least not dying immediately, but my only death in UT now are the rare occasions when a kobold crits me right outside of the first room and I cannot escape.

BTW, if you want challenge in early game, try old 0.50.9 where I messed up goblin spawn rates and they were appearing in UT. It was still survivable, but you had to be super careful if one spawned.

AquariusPower:
despite I like the increased challenge, this actually makes a lot of sense, 

> I probably should decrease the spawn rate of golems a bit to make up for their increased variety.

and I have been seeing a lot of golems, mostly weak ones, I just check to see what they are made from before fighting or fleeing xD

Script/char.dat
```diff
@@ -4047,16 +4256,57 @@ zombie
...
       "\"No one remembers Priscus, the former High Priest of Valpurus, these days. Only me.\"",
+      "\"The poor children of Khaz-zadm! Taken from their murdered parents and trained to die at the behest of the high priest, praising Valpurus with their dying breath.\"",
+      "\"Khaz-zadm once held the richest deposits of valpurian ore on this whole continent. No wonder Attnam has no shortage of vaplurium these days.\"",
+      "\"Attnam and Khaz-zadm were close partners for many years. Attnam bought its sacred metal from us, but we needed attnamese priests to bless the smelting, lest a burnt slag was produced instead of a magical metal.\"",
+      "\"I remember with fondness our shared visits with high priest Priscus. I was greatly saddened by the betrayal of the orcish delegation and his death. Too late did I learn the truth - I fell for the tricks of the same enemy he did.\"",
+      "\"High priest Priscus had a life-long dream - Pax Attnamica, a peace amongst all the lands, races and religions. Alas, not everyone even in his own country agreed with his pacifistic ways and evetually, his efforts led to his untimely demise.\"",
```
fejoa: 'eventually' instead of 'evetually'

Script/char.dat
```diff
@@ -4047,16 +4256,57 @@ zombie
...
+      "\"Many have recounted the story differently, but I believe this: It was neither the retinue of Priscus, nor the orcish delegation of Al-Khan who betrayed and massacred the other. There is someone else to blame.\"",
```
fejoa: Suggest 'Vul-Khan', or 'Vol-Khan' instead of 'Al-Khan'

red-kangaroo: I rather like Vul-Khan. It's now cannon that orcish assassin use a nerve pinch on their targets.

red-kangaroo: BTW, I wonder how many people will ever get to read this and similar dialogue. :D

red-kangaroo: > Shall we make this change?

Done. :)

PR #351, opened 2018-05-11
AutoPlay for Wizard mode to help on developer tests.

AquariusPower:
I added a autoplay feature, that uses existing NPC AI.
It is only available thru wizard mode.
To enable/disable it you need to press '~' key.

Every time we rest '.' it will not and will do some action like fight, eat, move etc...

I am trying to let it play by itself every 1 seconds or faster, so we can let it playing and do other things, waiting for crashes to fix.


AquariusPower:
I was thinking if this should be an external tool, not deployed/compiled in the main distribution package?
like an addon "dll" (I still have to learn how to make cpp plugins, may be interesting/useful).

fejoa:
If I use `~` in New Attnam for example, and then press `esc`, and then leave New Attnam, I get a crash.

Works awesome in the dungeon! It picked up a corpse of a mutant bunny and got stressed though... :D

AquariusPower:
mmm could be this!
```
DBGMSG:SIGNAL[11]='Segmentation fault'

./bin//ivan(_ZNK5level10GetLSquareEii+0x23)[0x6caa37]
./bin//ivan(_ZN9character39AutoPlayAICheckAreaLevelChangedAndResetEv+0xae)[0x6940e6]
./bin//ivan(_ZN9character26AutoPlayAIDebugDrawOverlayEv+0x2f)[0x694159]
./bin//ivan(_ZN4game20DrawEverythingNoBlitEb+0x7c8)[0x76e7be]
./bin//ivan(_ZN4game14DrawEverythingEv+0xe)[0x7658e0]
./bin//ivan(_ZN9character16GetPlayerCommandEv+0x34)[0x69b05c]
./bin//ivan(_ZN9character2BeEv+0x5ed)[0x68afd1]
./bin//ivan(_ZN4pool2BeEv+0x3b)[0x6827f3]
./bin//ivan(_ZN4game3RunEv+0x3da)[0x76550a]
./bin//ivan(main+0x475)[0x6f9d32]
```

autoplay was disabled, (after letting it cause havoc in new attnam) I went to dungeon and when trying to go to wilderness it crashed too.

when stressed, he will throw things everywhere furiously xD

fejoa:
Yep that sounds like it

AquariusPower:
as it is not intended for wilderness, but the AI may travel to it (and promptly go back to dungeon), I must prevent things that break on wilderness (like that fix, may be is was only that).

btw, travisci failed about stuff unrelated to the code again :>

PR #355, opened 2018-05-18
(ready) Show Map

AquariusPower:
EDIT: yep, been using it a lot w/o issues.

Mainly when using xBRZ x6,
when look mode is not helping to understand where we are.

It is actually the un-stretched vanilla dungeon displaced to fit in the window size.
(May be one day it could be the shrinked dungeon? tile 8x8 dots or even 4x4 dots :))

Also, it is simplified, will only show the dungeon, no mobs, no pets, only the player, no effects etc, just the terrain and terrain objects.

EDIT: it spends one turn as any other simple action, if you think it is cheap, it could spend more turns :)

EDIT: there is a new feature that lets another layer be drawn above the stretchedbuffer, it is where the map is drawn

DONE: optinally activate it during "detect material" mode

EDIT: 
Completely reworked it to actually look more like a map.
It is now as fast as possible.
It now has dynamic size based on currently discovered dungeon limits to not spoil what direction we should go, this means the map will shrink as farer you move (to fit the squares will get smaller).

image

The immersive mode shows where the player is standing (like at his hands) and is tiny (and is xBRZ also)
selection_092

PS.: later/one day we can play with it too: if player is confused, invert it, show things where there is none, and so on :)


fejoa:
This works awesomely! Thanks @AquariusPower

PR #359, opened 2018-05-22
Fix cure diseases when polymorphed

AquariusPower:
actually fix DeActivateTemporaryState() to also do it on the polymorph backup

but what we will see is when talking to the priest/god for a cure,
and player is still polymorphed in a werewolf,
after the polymorphing ends, lycantrophy (poison, leprosy etc based on what you paid for) will also all have been removed from the human form.

I think that was clearly a bug and not a difficulty thing, that looked like it was really missing right?


AquariusPower:
I also throwed in here this fix (commit log):
_added a fix, when a new game is started, 
when exiting and entering a dungeon subsequently many times
before saving by exiting to main menu at least once,
the CurrentSavefileVersion will be set to SAVE_FILE_VERSION on a new game._

PS.: if you perfer I can separate it in another PR.

fejoa:
>added a fix, when a new game is started,
when exiting and entering a dungeon subsequently many times
before saving by exiting to main menu at least once,
the CurrentSavefileVersion will be set to SAVE_FILE_VERSION on a new game.

Does this fix the bug you described in #342, about the crash on the levels with lions and bears, described here?:
>when it randomly happens, when going to gloomy caves the 1st time after playing a lot even at XT, I get this backtrace (about light emmiters, and this is about the 3rd time) I still dont know how to fix it!

It looks like a standalone bug fix to me

AquariusPower:
No, that fix for the CurrentSavefileVersion is just that the default value was -1 wainting to be set by a load game stack trace. The New game stack trace has no load therefore it was not being set.

that https://github.com/Attnam/ivan/pull/342 is about light emitters and I didnt looked that code in detail yet, mostly becase it is random, happens rarely and the new safe savegame is able to counter (woraround) it and probably other difficult situations.

Should I split CurrentSavefileVersion issue in another PR?

fejoa:
>Should I split CurrentSavefileVersion issue in another PR?

Nah, let's just merge

PR #360, opened 2018-05-22
(needs testing) New materials

red-kangaroo:
Add many, many new materials. There are now multiple hardening lines (eg. bronze, line, iron line, cloth line, feater line, ...) and several non-metal weapon grade materials.

[Here](https://attnam.com/topics/New-materials) is the forum thread and [here](https://pastebin.com/raw/7d7xt3vi) is the wizmode list of materials.

Any testing is more than welcome. I tried my best to balance the materials against the current ones, but if you find some Int requirement or something weird, please tell me.

Material chances for most items have yet to be added (only weapons are done so far).


red-kangaroo:
Damn, it seems I have failed to properly split this PR and New Monsters. This PR starts at [material manual](https://github.com/Attnam/ivan/pull/360/commits/19178a58641742e2757d5a81d9de49deb54e3b67) commit.

AquariusPower:
You can leave this PR as depending on the other 1st, may be?

Btw,

I had about 3 separated simultaneous working branches. To work on each, change between them, compile, merge it all etc etc etc... was a huge complicated time consuming mess...

So I ended creating these git tools:
https://github.com/AquariusPower/CppDebugMessages/blob/master/.devsPrefs/AquariusPower/gitConfigMultiBranchTools.txt
1st - create the .sbkp file (it is a simple copy, you should now backup everything, I use Unison)
1.1 - revert the file (.cpp or .h) so you can prepare a "relative patch"
2 - create the "relative patches"
2.1 - change to the branch you want to apply the patch
3 - apply the "relative patch" and resolve minor conflicts
(at that same path on that repo, there are some command line tools .sh scripts to let it work on many files at the same time, as thru git gui we can only work in one file per time).

after using it, it is now a breeze (as possible) to prepare the work for many branches in a single merged branch, and split all the new "relative code" to other branches.

Obs.: it is all for **linux** (and probably Mac OSX too I guess). But I think they are not complicated to be ported to windows as probably only a few commands will change if at all.

PS.: I still have to stackexchange this and may be add them as dev tools here too...

fejoa:
Awesome tool @AquariusPower. Do you have a gist/post describing how to use it?

AquariusPower:
@fejoa well, I should go to stackexchange to add more details in a clearer way and see if other ppl can improve it. So the best we have for now is this: https://github.com/Attnam/ivan/pull/360#issuecomment-391159583 :wink:, let me try add some more clearer details.

red-kangaroo:
@AquariusPower SOHM can already harden up to the highest material in a hardening chain. I'm not exactly sure about the mechanics, but try getting some higher Int and hardening eg. balsa spear. It will end up kauri or so, jumping over all the intermediate woods. It's only with low Int that you need to go one tier at a time.

AquariusPower:
so, as I understand, it actually makes SOHM works better (more precisely), interesting, thx on the info! indeed, Int 10 wont help much I guess lol

fejoa:
Just posting a link to a diff for convenience:
https://github.com/red-kangaroo/ivan/compare/NewMonsters...NewMaterials

red-kangaroo:
@fejoa Would it be possible to compile a Windows testing build? With the amount of added stuff, I would like to get as much feedback as possible, and a build could be put eg. on Discord to have some players try it out.

AquariusPower:
@red-kangaroo I was taking a look at the failed compilations
moving at proto.h this line to the `#ifdef WIZARD` above will probably fix it
`static void CreateEveryMaterial(std::vector<material*>&, const god*, ccharacter*);`

@fejoa the linux [git tools](https://github.com/AquariusPower/CppDebugMessages/blob/master/.devsPrefs/AquariusPower/git tools/readme.txt) are working well now (actually the most useful 2nd part that process many files at once),
if you want to use them on windows, I suggest [cygwin](https://cygwin.com/install.html)

fejoa:
@red-kangaroo here it is, can you test? https://github.com/Attnam/ivan/releases/tag/v053

red-kangaroo:
@fejoa Everything seems to work just fine on Windows, I'll put it out on forum and Discord. Thanks! :)

Issue #362, opened 2018-05-24
random fantasy name generator


AquariusPower:
https://github.com/skeeto/fantasyname

when starting a new game.
an option to auto fill it.
if you dont like the name, just go back to main menu and hit start new game again.
looks quite simple to implement (and will prevent all the messy names tha at least I create during tests lol).

red-kangaroo:
That's a great idea.

red-kangaroo:
Already done!

PR #363, opened 2018-05-26
(ready) Fantasy name generator

AquariusPower:
https://github.com/Attnam/ivan/issues/362
all documentation for the pattern is at `namegen.h`
EDIT: btw, they use "unlicense"


AquariusPower:
what about use this to also give random names to mobs? 
simplified ones, like only having the 1st name. 

but for ex. kobolds could have some fixed name part, and so on to all other NPC types, but I am not good on naming and history/story and so on, so tips on this are welcome.

fejoa:
I can't seem to get this working, do you have example screenshots for how to configure? 
I set the pattern to `!ss !sV` as suggested, however when I start a new character, the name field doesn't populate with the randomly generated name. I closed the application and re-opened and still no name.

AquariusPower:
if you have configured a default player name, it will override the random name, or is it coming empty?

fejoa:
With no default name configured and with fantasy name pattern `!ss !sV`, when I start a new game, I get prompted for a new name and the field is empty.

fejoa:
[windows] I compiled namegen.exe and ran `./namegen "!ss !sV" 5` from PowerShell, which gave me the following output
```
>combinations = 33459250





>
```

So I guess the five carriage returns are where the strings are supposed to be? It must be a difference in the way the namegen strings are typed/stored on UNIX vs windows???

Sorry, I couldn't get the debug tool working :(

AquariusPower:
:o
here how it looks like:
```
> combinations = 33459250
Therat Rilie
Estir Rodia
Enthet Hine
Kalorm Asou
Nalight Asy
Therlye Irou
Onkim Oldea
Alewor Tanai
Hintin Ormay
Lyetor Garie
```
it is failing for windows for some reason then :/
I will take a look on it

AquariusPower:
the readme says "It requires C++11."

my compilers (g++ c++ cpp gcc ..) are version 5.4.0

this code
```
  if( __cplusplus == 201103L ) std::cout << "C++11\n" ;
  else if( __cplusplus == 199711L ) std::cout << "C++98\n" ;
  else std::cout << "pre-standard C++\n" ;
```
outputs C++98 (so despite it requires C++11 it works with an older version?)

AquariusPower:
the fix @skeeto provided (thx!) still works on linux, 
@fejoa could u confirm for windows?

fejoa:
It works! Thanks @AquariusPower and @skeeto !

For later:
- Love to be able to bring some of these lines into the datafiles, so users can customise without compiling?
- Perhaps strings/names in the namespace can be attributed to characters in `char.dat`, so that bears can have bear-sounding names, goblins goblin-sounding names etc?
- Make names discoverable by sci-talking, charming creatures with above certain int... etc? 
- Use willpower checks to allow changing pet names?

AquariusPower:
animal friend skill: discovering the real animal name and calling it could be another way to taming!

makes me think about: may be one or two powers/perks/traits could be randomized when the player is created, and we could use them like we use the scrolls and wands, once an in-game day, a non cumulative applyable ethereal rune would spawn in our inventory for that to work using current game mechanics. (mmm.. this feels like a issue text..)

so, you mean like using the namegenerator pattern in char.dat? we just need a token like start the name with '#' and it would the generator instead of fixed name.

and if changing the pet name when we are confuse, the name gets randomly scrambled :>

PR #366, opened 2018-05-27
Fix backspace key on MacOS & Support Home key and End key

iology:
A recent change broke backspace key on OS X.

### DIFF
--- a/FeLib/Include/feio.h
+++ b/FeLib/Include/feio.h
@@ -43,7 +43,7 @@ class iosystem
   static void TextScreen(cfestring&, v2 Disp = ZERO_V2,
                          col16 = 0xFFFF, truth = true,
                          truth = true, bitmapeditor = 0);
-  static truth IsAcceptableForStringQuestion(char);
+  static truth IsAcceptableForStringQuestion(unsigned int);
   static truth IsOnMenu();
 };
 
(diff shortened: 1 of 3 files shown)


AquariusPower:
working on linux too!

fejoa:
It breaks left and right key behaviour in question fields on windows

AquariusPower:
I guess was this:

-  static truth IsAcceptableForStringQuestion(char);
 +  static truth IsAcceptableForStringQuestion(unsigned int);

may be convert to `char` for some of the tests could fix?

fejoa:
> may be convert to char for some of the tests could fix?

yep switching back to `static truth IsAcceptableForStringQuestion(char);` worked. I guess it just needs to be for left and right on windows. I don't know of a good C conversion between uint and char...

iology:
The signedness of `char` can vary. How about changing it to `static truth IsAcceptableForStringQuestion(int);`? The values of KEY_LEFT/RIGHT/HOME/END are bigger than 256 (what a 8-bit char can hold). Or we can move the check for control codes and key codes out of IsAcceptableForStringQuestion().

iology:
The C standard library uses `isdigit(int c)`, so `int` should be okay. It is lucky that the key codes are also in the range of 0 to INT_MAX.

iology:
Ah, I haven't included KEY_LEFT and KEY_RIGHT in `IsAcceptableForStringQuestion()`, just added.

PR #370, opened 2018-05-29
(ready) Add item label

AquariusPower:
Now the user can add a label to any item (or amount of equal items), they will stack if having same label (or none as vanilla).
selection_094

should it be called something else than "label"?

as you can see, I didnt force any delimiters (should I?) to let user decide how it should look.

it is prepended, should be elsewhere?

should be more immersive? (use a dagger to engrave the item (hard floors should require it too btw)? :), trying to engrave a wand may make it explode etc. Should it give exp to int or wisdom?)

for now it is just useful:
- an empty wand
- items to sell
- something to be kept for special event
- a small chest that belongs to pet inventory
- a small chest only for magical items

### DIFF
--- a/Main/Source/command.cpp
+++ b/Main/Source/command.cpp
@@ -76,11 +76,12 @@ command* commandsystem::Command[] =
   new command(&Go, "go", 'g', 'g', 'g', false),
   new command(&GoDown, "go down/enter area", '>', '>', '>', true),
   new command(&GoUp, "go up", '<', '<', '<', true),
+  new command(&SetItemLabel, "inscribe item", 'b', 'b', 'b', true),
   new command(&IssueCommand, "issue command(s) to team member(s)", 'I', 'I', 'I', false),
   new command(&Kick, "kick", 'k', 'K', 'K', false),
   new command(&Look, "look", 'l', 'L', 'L', true),
   new command(&ShowMap, "show map", 'm', 'm', 'm', false),
-  new command(&AssignName , "name"     , 'n', 'n', 'N', false),
+  new command(&AssignName , "name", 'n', 'n', 'N', false),
   new command(&Offer, "offer", 'O', 'f', 'O', false),
   new command(&Open, "open", 'o', 'O', 'o', false),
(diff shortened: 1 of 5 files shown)


red-kangaroo:
This is great!

> should it be called something else than "label"?

I thin label is OK, but if you want some options, maybe inscription, engraving, called or named...

> should be more immersive?

Easy labels are OK, IMHO, just as the player can easily engrave with nothing in their hands.

Main/Source/command.cpp
```diff
@@ -503,13 +503,13 @@ truth commandsystem::SetItemLabel(character* Char)
...
     game::DrawEverythingNoBlit();
-    Char->GetStack()->DrawContents(ToAddLabel, Char, CONST_S("What do you want to add a label?"), REMEMBER_SELECTED);
+    Char->GetStack()->DrawContents(ToAddLabel, Char, CONST_S("What do you want to inscribe?"), REMEMBER_SELECTED);
 
     if(ToAddLabel.empty())
       break;
 
     festring What = ToAddLabel[0]->GetLabel();
-    if(game::StringQuestion(What, CONST_S("How you want to label it?"), WHITE, 0, 20, true) == NORMAL_EXIT)
+    if(game::StringQuestion(What, CONST_S("What would you like to inscribe in this item?"), WHITE, 0, 20, true) == NORMAL_EXIT)
```
red-kangaroo: inscribe ON this item :)

AquariusPower: thx! xD

AquariusPower: added some more on(s), not sure if all correct tho

iology:
So many crazy ideas. What if you lost memories? Items burnt?

AquariusPower:
@iology 

these can be new features one day, feel free to code them!  :)
the inscription could be blurred like

!KeepThis 

could become 

!K3epT415 (blurred if it gets wet), 
#K####hi# (burnt or dirty),

as with the map, if we are confused, could show wrong indicators or even change the displayed dungeon  path.

The initial goal is to add useful features. 
Immersion may come later :)

PR #372, opened 2018-05-29
(ready) GoOn can now ignore items

AquariusPower:
by usability
cheap
encumbering (heavy and cheap)

(this may remove the challenge of not finding useful things (under other criteria) in the messy battle grounds, if u think so, just disable it)


AquariusPower:
The cheap things to be ignored use a max price of 5.
The encumbering is checked against 10.
I think these may be better tweaked later, or it would be better become a user option?
The doubt on what to keep to be sold is also a challenge, and letting it be a user option could lower such challenge, not sure tho...

I will set it ready, unless you prefer that I add such option. The good point on adding it is that someone may figure out a default value based on tests tho :)

Issue #357, opened 2018-05-22
Provide an option to always use nearest-neighbor scaling for upscaling?


iology:
From the source code we can see that nearest-neighbor scaling is always preferred when possible:

https://github.com/Attnam/ivan/blob/3d1271d234e8ebdc796289e94ca1ff8b36849271/FeLib/Source/graphics.cpp#L223-L250

Without an extra large (external) HD display, linear scaling currently only makes more sense for downscaling, for upscaling it blurs the picture **especially for the text using a bitmap font**. There are no higher-resolution image/font assets for downsampling. e.g. 2x for around(800x600) < size < around(1600x1200)

I can't stand the blurred text on my retina-mabook, so I always disable linear scaling. The pixelated text hurts much less than the blurred text does for my eyes.

AquariusPower:
may be the font can be upscaled 2x, 3x.. 6x using xBRZ, that file saved and used as alternative, so no need to draw manually, but I am not sure if the result will be usable as a font, some tests are required (if I clearly understood the problem)

iology:
The font issue is indeed foremost to me. But the images are also scaled, so even the 1x dungeon looks different when using different scaling algorithms, however the blurring is less obvious with 3x-6x dungeon. I use 3x dungeon. My situation is (not in full screen mode):

* `VirtualDisplayMode` (virtual screen size of the OS): 1280x800 (actual: 2560x1600), device-pixel ratio=2
* `NewRes` (virtual screen size of the game): 800x600
* `ActualWindowRes` (drawable canvas size for the game): 1600x1200
* `ActualDisplayRes`: 2560x1600 (not considering the top bar of MacOS and title bar of the game window ;-)

I feel the logic in the code is wrong, will delve into it later. According to the result, the picture is scaled. 2x, 4x, 6x, 8x, 10x, 12x needed for downsampling to 1x-6x dungeons and text (because the pixel ratio is 2.0, not 1.5, 3.0, ...)

Anyways, I like pixel art and my eyesight can bear with the dense but still legible pixelated text (in nearest-neighbor scaling mode). My patch is simple: just `SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, "nearest")`. :P

<details>

<summary>Some screenshots (if you have a rMBP to check them):</summary>

1x dungeon (linear):
[...]

PR #373, opened 2018-05-30
Add a graphics option for nearest-neighbor and linear interpolation

iology:
Ping #357. Currently there are "nearest-neighbor" (use British English "neighbour"?) and "linear".

opt-scaling-quality

But the "nearest-neighbor" is too long for in-game menu, so it goes to the next line. Suggestion?

opt-scaling-quality-playing

p.s. It seems the code for DOSBox is not maintained anymore, should we remove all related code?

### DIFF
--- a/FeLib/Include/graphics.h
+++ b/FeLib/Include/graphics.h
@@ -41,13 +41,14 @@ class graphics
 #ifdef USE_SDL
   static void SetScale(int);
   static void SwitchMode();
+  static void SetMode(cchar*, cchar*, v2, int, int, truth);
 #endif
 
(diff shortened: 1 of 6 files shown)


iology:
Alternatives: nearest&linear pixelated&smooth

fejoa:
Nearest and linear sound most accurate, but pixelated and smooth are more descriptive. Maybe opt for the latter. I'm tempted to call linear or smooth, "fuzzy".

iology:
I prefer pixelated&sm👀th too. Rebased and check passed!

PR #375, opened 2018-05-30
Savegame list fix and filename changes

AquariusPower:
I emptied my Save path.
Created a new char and played thru some dungeons w/o saving to main menu.
As I was also testing thing, it crashed, and I ended up only with .bkp and .AutoSave.sav files, but the main PlayerName.sav was missing.
The list would not be populated.

To test it, do the same but crash it manually by killing the game process.

I also hardened the filenaming to only allow a-zA-Z0-9 preventing some derivative problems if not.

This one will be ready after we make many tests and find no troubles in:
- the savegame filenames
- the player name

To properly show the player's name on the savegame list, we need to create a new file during the savegame, it can be a simple text file like PlayerName.info may be, and there can be put many things, including the player name with all fancy characters and also current game/dungeon progress, etc. Not need to be on this PR, that is more a bug fix and code hardening.

### DIFF
--- a/FeLib/Source/feio.cpp
+++ b/FeLib/Source/feio.cpp
@@ -933,32 +933,37 @@ struct fileInfo{
   char* GameScript = 0; //dummy
   int CurrentDungeonIndex = -1;
   int CurrentLevelIndex = -1;
-  festring name=festring(); //TODO this init helps with festring? it is buggy?
-  festring fullName=festring();
+  festring fileName=festring(); //TODO this init helps with festring? is it buggy?
+  festring absFileName=festring(); //contains the full path
   festring time=festring();
   festring idOnList=festring();
   festring dungeonID=festring();
-  festring nameAutoSave=festring();
+  festring fileNameAutoSave=festring();
   std::vector<festring> vBackups;
   bool bIsBkp = false;
 };
 std::vector<fileInfo> vFiles;
-bool addFileInfo(char* c){
+bool addFileInfo(const char* c){
   // skippers
   if(strcmp(c,".")==0)return false;
   if(strcmp(c,"..")==0)return false;
 
   // do add
   fileInfo fi;
-  fi.name<<c; //TODO this assigning helps with festring instead of '=', it is buggy?
+  fi.fileName<<c; //TODO this assigning helps with festring instead of '=', it is buggy?
   vFiles.push_back(fi); //stores a copy
 
   return true; //added
 }
 skipseeksave skipSeek=NULL;
 void iosystem::SetSkipSeekSave(skipseeksave sss){skipSeek = sss;}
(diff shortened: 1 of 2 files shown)


fejoa:
Sorry for another possible bug. If I load an autosave file, the character's name is missing.

AquariusPower:
@fejoa actually if as many bugs as possible are found before release is a great thing :)
As I remember, I had solved it from a normal .sav file, but didnt check from .AutoSave.sav file.
So I thx for finding it! :)

fejoa:
Shall I merge this and .AutoSave.sav can be a new job?

AquariusPower:
sure! :>

AquariusPower:
mmm... I am not being able to recreate the problem.
I let it autosave (25 turns here), and killed the application.
One of the times there was only autosaves, the other it had also .bkp files.
but in both the player's name was ok.
even in wizard mode that doesnt delete autosave files.

btw, wizard mode has a problem about that.
wizard autosave files are prefered just because they exist, even if the .sav main file was saved later,
this means that when we make tests in wiz mode and save, when we continue it will be in the past (in the last autosave turn), I didnt fix that yet...
the savegame list just blindly prefers the autosave.sav as that is the normal for non-wizard mode.
and I dont know how to address that yet... only the datetime of the .sav files could suffice, but best would be to read the current turn value from each and the higher one wins, not sure tho yet.

so, if you still have that savegame and could upload here it would help!
otherwise we wait it happen again.

red-kangaroo:
@AquariusPower  I had a similar problem. Game crashed and I tried to recoved from savefile. There was an autosave, but for whatever reason the game instead tried to use normal savefile. I didn't save before, though, so it threw me back to the story at the very beginning. When the game started, I had no name.

null

Is there a way to tell the game to recover from an autosave and not the normal save? I would rather not loose a lot of progress if the game crashes and instead of going back to the nearest autosave, it throws me all the way back to the last regular save.

AquariusPower:
@red-kangaroo 
>I didn't save before, though, so it threw me back to the story at the very beginning. When the game started, I had no name.
as I remember, I had this issue and fixed it, so I guess this was from a previous commit than the latest I did on that branch.
A good way is to check the savegame (or better the ivan built executable) file date-time, so we can compare with the commit date-time and be sure it was previous to the latest ones.

so in case you still have such savegames, it would be interesting to paste here the file list,
as I think that "1    0     " could be actually a dungeon level ID, therefore your savegame may be something like "playerName.10", but I need to be sure because there is spaces between 1 and 0 and it would be a lot of guess work w/o having a proper stacktrace/backtrace when the crash happened.

speaking of that... we seriously need to find a way to generate stacktraces for the windows build.
Even if you enable the debug log, it only shows things where debug entries were added and will not help on random crashes.

I am trying to find a post I did about windows stacktraces on stackexchange, brb

PR #376, opened 2018-05-31
Adjust the key for inscription for the NetHack scheme

iology:
'b' is already used for going southwest. Let's make it 'B'.


iology:
btw can we combine engraving and inscription?

fejoa:
> btw can we combine engraving and inscription?

Yes I think inscribing should use the same command key as engrave (`G`), only a question should come up like:
`Select inventory (i) or square (.). Press esc to abort.`

Or something to that effect.

AquariusPower:
@fejoa yes! that is really good, initially I thought on doing that but couldnt figure out how.

Issue #377, opened 2018-05-31
Combine command for engraving and inscription


fejoa:
See #376 
https://github.com/Attnam/ivan/pull/376#issuecomment-393396472

fejoa:
closed by #378

PR #378, opened 2018-05-31
Combine command for engraving and inscription

iology:
#377 #376 Illiterate cannot engrave nor inscribe. 😛


iology:
@fejoa ah, I just rebased and your commit was lost. :P

PR #385, opened 2018-06-01
bugfix for MaxValueless duplicity

AquariusPower:


AquariusPower:
np, as I didnt know if u were online, I hurried to fix it as was from my commits (I used it in 2 diff branches), but was good you had fixed it 1st :)

Issue #379, opened 2018-06-01
confdef.h and define.dat are not in sync


AquariusPower:
as you can see on the picture (wizard mode),
![selection_095](https://user-images.githubusercontent.com/17087075/40819490-bafcabb2-6531-11e8-9ae4-9e3faad8c8b3.jpg)
there are 2 missplaced chairs created by the eddys,
and a couch created by them too.

define.dat
```
#define COUCH 5
#define CHAIR 6
#define DOUBLE_BED 9

```
confdef.h
```
#define COUCH 5
#define DOUBLE_BED 6
```

eddy code:
```
void eddy::GetAICommand() ...
    decoration* Couch = decoration::Spawn(RAND_N(5) ? COUCH : DOUBLE_BED);

```

as confdef.h and define.dat are not in sync, 
eddy should have spawned config value 9 from define.dat as the identifier is DOUBLE_BED (the intended object),
but it spawned a chair (6) the out of sync configuration.

Considering define.dat code is just c++ code, 
it should be used as such, overriding all defines from confdef.h,
to prevent double maintenance (that led to this out of sync unintended result).

So, `Script/define.dat`, during compilation, could be copied to `Main/Include/define.h`, 
overwritting it there.
OR
during compilation, `Script/define.dat` be compared to `Main/Include/define.h`, 
and fail, saying they are out of sync, if not equal, 
what is better to prevent losing work if you code it at Main/Include/define.h.

or am I wrong? :)

PS.: I will create a script to check both files to see the extension of the out of syncness.

AquariusPower:
here is the sorted '#define' results from both files:
[confdef.h.log](https://github.com/Attnam/ivan/files/2060898/confdef.h.log)
[define.dat.log](https://github.com/Attnam/ivan/files/2060899/define.dat.log)

When comparing them, you will see that there are a lot of identical entries.
But all NEW entries are from `define.dat`.

And there are many entries out of sync (possibly causing unpredictable (now predictable) results):
```
> #define ANVIL 12
> #define BIRCH 16
> #define CACTUS 14
> #define DOUBLE_BED 6
> #define DWARF_BIRCH 18
> #define FLESH_ID (5 << 12)
> #define GAS_ID (3 << 12)
> #define IRON_ALLOY_ID (7 << 12)
> #define LIQUID_ID (4 << 12)
> #define OAK 15
> #define ORGANIC_ID (2 << 12)
> #define PALM 9
> #define POOL_BORDER 7
[...]

PR #387, opened 2018-06-01
(ready) define.dat validator: dynamic code generator

AquariusPower:
more details and latest updates here: https://github.com/Attnam/ivan/issues/379

I saw no problems on initial tests,
I am not used to trees visual differences tho, but saw nothing weird either.

tests are welcome.

the validator dynamic code generation and actually perform the validation must be triggered by in-game config option, and will abort if something mismatching is found.


fejoa:
I compiled it and ran, it seems stable. What else should I do to test the functionality of this?

AquariusPower:
to see how it works, modify the define.dat file like
1) change ex. TORSO value to 2 (so that it will differ from TORSO value at ivandef.h, could be other .h file tho), 
actually the problem will happen when someone create a new value at define.dat and use another value at c++ .h files with the same `#define` identifier,
or when someone changes a value only at `define.dat` or at some c++ .h file, so they we be out of sync.
2) compile and install the game.
3) go to that option and toggle it to yes.
4) the file `definesvalidator.h` will be created at the same path where the `ivan.conf` is located.
5) move `definesvalidator.h` to the project development path at `Main/Include/definesvalidator.h` overwriting it
6) compile and install again so that the validator will know what value the c++ `#define` should have, based on the define.dat file.
in case c++ has no such define, it will be ignored ex.: the check will be performed only `#ifdef EMISSARY` (currently not used in c++ .h files)
7) run and activate that option to `yes`, it will abort complaining about TORSO value with this message: "Defined TORSO with value 2 from .dat file mismatches hardcoded c++ define value of 1!"

I will add the text above to a doc.
Anyone can do it anytime, and the inconsistencies will show up.
If changing values in define.dat or c++ files and do not do these steps, 
the problem will remain unnoticed, and eddy may drop something else other than double bed again :)
So, to do it is not necessary, but if anyone wants to check for consistencies, it will work.

And... if there is a better way than following all these steps... it will be mostly welcome fore sure! :D

PR #390, opened 2018-06-03
More interesting final TX level

red-kangaroo:
Resurrected Xinroch's power is fine, but IMHO his level is a bit
lacking. Compare with Oree's lair, which has a design unseen anywhere
else in the game.

I suggest using a reversed TX Entry temple, ie. a reversed pentagram
where Xinroch can stand and fight. Fake UT1 entry for a bit of fun with
the player, then a fight in open area around the temple, and finally a
showdown with Xinroch in ruby-and-diamond temple.

With the new materials at our disposal, I changed the earth and wall
materials for Xinroch ghost and resurrected Xinroch levels a bit to have
a nice color scheme. :)

### DIFF
--- a/Script/dungeons/Attnam.dat
+++ b/Script/dungeons/Attnam.dat
@@ -22,7 +22,7 @@
 Dungeon ATTNAM;
 {
   Levels = 5;
-  Description = "Attnamese Catacombs";
+  Description = "attnamese catacombs";
   ShortDescription = "AC";
 
   LevelDefault
@@ -84,8 +84,8 @@ Dungeon ATTNAM;
 
   Level 0;
   {
-    Description = Attnam;
-    ShortDescription = Attnam;
+    Description = "Attnam";
+    ShortDescription = "Attnam";
     FillSquare = solidterrain(SNOW_TERRAIN), 0;
     Size = 61, 67;
     GenerateMonsters = false;
@@ -878,8 +878,8 @@ Dungeon ATTNAM;
 
   Level 1;
   {
-    Description = Cellar;
-    ShortDescription = Cellar;
+    Description = "cathedral cellar";
+    ShortDescription = "Cellar";
     Size = 42, 48;
     GenerateMonsters = false;
     Rooms = 1;
@@ -1342,6 +1342,8 @@ Dungeon ATTNAM;
 
   Level 4;
   {
+    Description = "lost tomb";
+    ShortDescription = "Lost Tomb";
     LevelMessage = "You hear rhythmic chanting of some profane ritual: \"Ia! Ia! Khaz-zadm fhtagn!\"";
     DifficultyBase = 100;
     DifficultyDelta = 0;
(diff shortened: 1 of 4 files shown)


fejoa:
Hmm, TX already suffers from plenty of design flaws, admittedly largely by my own fair hand. The level looks impossible to beat in this change. I guess the Cathedral is much the same difficulty; an open level with plenty of rugged hostiles. Perhaps we should reduce the number of dark knights by 5 or 10? Otherwise it's just drudgery hacking limbs and such. I guess we can only merge and find out from the power players what their feedback is.

I think the shadow veil thing might be a bit clunky. There's still that bug where it can be dropped in the wall, so maybe that needs to appear in the player's inventory instead?

Need to think of more ways in which to make this an enjoyable experience for players. TX is just vicious, and even though I instigated its development, it's not a dungeon that I personally play in very often. 

Anyway, I'm happy to merge this, as things can always be tweaked later. Any other comments from anyone else?

red-kangaroo:
> Hmm, TX already suffers from plenty of design flaws, admittedly largely by my own fair hand.
> TX is just vicious, and even though I instigated its development, it's not a dungeon that I personally play in very often.

Hey, I actually play TX more often that GC! The reason I'm proposing various changes to TX is because I like the dungeon and want to see more of it, not because I think it's bad. :)

It might be vicious, but in my experience, vicious in a good way.

> Perhaps we should reduce the number of dark knights by 5 or 10? Otherwise it's just drudgery hacking limbs and such. 

Actually, here I'm counting on the shadow veil. Oree's lair is also a level you don't really want to slowly clear. There's too much risk. With shadow veil and many, many enemies, the player will have to get creative, dodge into walls, pick his fights and most probably just head straight for Xinroch. Starting a killfest will get you swarmed and killed quickly, but that's the point.

Maybe that's a wrong design? I could definitely decrease the number of enemies, if you think that's appropriate?

> I think the shadow veil thing might be a bit clunky. There's still that bug where it can be dropped in the wall, so maybe that needs to appear in the player's inventory instead?

IDK, it's easy enough to avoid getting it dropped into a wall, plus there are so many ways of digging it out I don't see this as a problem. I might be a weird type of gamer, but I like getting punished for a lack of foresight - it's kinda obvious the loot will end up in a wall, so do something about it before it gets stuck ina  wall.

fejoa:
> Actually, here I'm counting on the shadow veil. Oree's lair is also a level you don't really want to slowly clear. There's too much risk. With shadow veil and many, many enemies, the player will have to get creative, dodge into walls, pick his fights and most probably just head straight for Xinroch.

>Starting a killfest will get you swarmed and killed quickly, but that's the point. Look at the welcoming party in Oree's lair - there's no point, only danger in fighting them.

>Maybe that's a wrong design? I could definitely decrease the number of enemies, if you think that's appropriate?

You have a good point, the veil is a decent piece of ascension kit for this level, which should handily help players along. Let's merge this : )

PR #391, opened 2018-06-03
(merge before release) Balance patch for the new materials

red-kangaroo:


AquariusPower:
this should be merged b4 they create the v0.54 tag?

red-kangaroo:
Yeah, before the release whatever its number should be. It's some minor balancing for the new materials added in my previous PR.

PR #393, opened 2018-06-06
(ready) improved felist key buffer usage to stop when releasing key up/down

AquariusPower:
Problem most visible when you got low fps.
On a felist like drop item, with scaled silhouette enabled (what will lower fps), 
if you hold key up/down it may keep processing that buffered key for about 3 seconds!! 
what could turn one or 2 pages more of items, quite annoying...

now, as soon you release the navigation key up/down it will promptly stop processing the buffer.

obs.: also made it more clear the sfx option text.


AquariusPower:
wait,
it randomly seems to stop animations waiting we to press a key during felist,
still trying to determine how that happens...

PR #396, opened 2018-06-08
Unlock chests that spawn broken locks

fejoa:
I noticed that chests with broken locks sometimes spawn locked. As there is no key to open a broken lock, I thought maybe they should be openable, meaning the player doesn't have to bash the broken lock open. 

Should chests with broken locks be jammed (current)? Or should they always be openable (this change)?

Doors can still spawn locked broken locks, and this change doesn't affect that behaviour.

### DIFF
--- a/Main/Source/miscitem.cpp
+++ b/Main/Source/miscitem.cpp
@@ -1278,6 +1278,10 @@ void itemcontainer::PostConstruct()
 {
   lockableitem::PostConstruct();
   SetIsLocked(RAND_N(3));
+  
+  if((GetConfig()&LOCK_BITS)&BROKEN_LOCK)
+    SetIsLocked(false);
+  
   long ItemNumber = RAND() % (GetMaxGeneratedContainedItems() + 1);
 
   for(int c = 0; c < ItemNumber; ++c)


AquariusPower:
I think both chests and doors could spawn (written when we look at them as) "jammed".
But I also think this fix is good because a "locked broken lock" is weird...

iology:
Why not fix the message instead?

iology:
Oh Frog, bananas have `Jammed` state: https://github.com/Attnam/ivan/blob/5e19f1d33ff1d99bd97e4d1bf790554bd7fcee7b/Main/Include/miscitem.h#L58-L75

iology:
We can provide `IsJammed() = locked + broken`. I think kicking every chest is fun. :P

fejoa:
Until you find a gas grenade inside. I like your jammed = locked+broken idea @iology

AquariusPower:
just to make it clear:
When kicking a locked lock it breaks and becomes unlocked.
Instead of creating jammed state it would be: **spawn** locked and broken means jammed (the proposed renaming).

But if it gets jammed after spawning? we could by mistake use a triangular key on a squared lock and jam! 
Or later, mechanics to let it "melt" by fire (or acid?) and gets jammed again.
Rusted by water too: jam!
 :)

fejoa:
Can we merge this as an interim solution?

AquariusPower:
sure! 
most stuff I talk are issues to develop in the future on other PRs anyway hehe

iology:
No problem. I'm not working on such things yet, which can spoil my fun. :P

PR #401, opened 2018-06-11
(ready) Quick swap weapons and hotkey

AquariusPower:
'b' swaps, 'B' configurations (**should we change these keys?** I liked cuz all are within easy reach from left hand, and all other appropriate keys were in use already)

You can have as many configurations you want.
**I didn't add checks for dup cfgs** (what would complicate the code), but you can easily remove them.

finally learned how to create felist properly with icons :) ~~Obs.: I didnt manage to show weapons icons w/o crashing, so I disabled the related coded lines. The selectable entries have no icons and the non selectable (showing weapons) should have icons.~~

It is working on initial tests.
I think there may have minor things that could be improved, not sure tho.
~~The missing item graphics is not cool, but the functionality is quite helpful, so when we manage to show the item's graphics it will be great, but not essential.~~
If there is nothing breaking the functionality, then it is ready.

EDIT: this will also make old savegames incompatible,
as we are still in v132, import is not possible from previous builds also at v132,
(unless I hack it to optionally force a "load-ignore" (for this new data) even if the version is already 132, but that sounds messy :/ )
I see this like: when it gets released, being v132, then new additions to the savegame data may be importable in v133 (one day) :)

Tests required:
- will cause problems if item was offered (destroyed) to a god?
- will cause problems after throwing or dropping the item? 
- will cause problems if item breaks after being assigned?

PS.: one day this could be extended to all equipments also.
PS.2:  fun fact: it took me a week to figure out how to implement this, and today I woke up with the answer xD


red-kangaroo:
This sounds nice. BTW, thinking about weapons and QOL features, is it possible to have autopick for weapons you have thrown, so that you don't need to pick them one at a time?

AquariusPower:
**Another scratch thought:**

I also thought about quick throw auto inscribe option.
May be to make that automatic, we could use the label (inscribe) to automatically add tokens to items.
So, to make it as easy as possible, every thrown item (that has no inscription already?) would receive a `#thrown` token.

I thought on the 'T' key. 
But actually, "throw (quick)" could be 't' (to not have to hold shift key), 
and "throw" (existing one) would now be 'T'.

How quick throw would work?
- if you do not move, it will throw in the last direction you chose, all items with `#thown` inscription.
- if you move, it will ask for a direction to throw the inscribed ones.
- if there is no inscribed, it will fallback to default throwing mode.
- if you want to throw an specific item, just use the vanilla throw mode.
- if you want to avoid re-trowing something, clear the inscription. but as this may be annoying, this inscription `#no-throw` could be used, so whenever it is thrown, the `#thrown` would not be automatically added.

Doesnt seem complex to implement, just need to know if that flow is good?

**autopick for weapons you have thrown**

so, the thrown with `#thrown` inscription could also be optionally auto picked up.

iology:
Please save the nethack direction key. D-: How about "x" and "X"? Wizard mode can use "`" instead.

iology:
Like NetHack, `#command` is good for advanced settings and infrequent commands.

AquariusPower:
> Please save the nethack direction key. D-: How about "x" and "X"? Wizard mode can use "`" instead.
I initially thought on x X, didnt chose cuz of wizard, but.. as "~" is wizard autoplay, "`" (same physical key) will be great for wizard!
@fejoa what do you think?

> Like NetHack, #command is good for advanced settings and infrequent commands.
`#cmd=filter:+wand,-potion` could even be a regex based on full item name! but would require advanced knowledge, so may be keep it simple? 
`#cmd=throwable`

AquariusPower:
@iology 
as this is not the 1st time I chose wrong keys
I think I will just add a check in case any of these are chosen for commands to promptly abort during commands config to alert developers:
game::MoveNormalCommandKey
game::MoveAbnormalCommandKey
game::MoveNetHackCommandKey

iology:
Oh, I meant `#<command name here>` (e.g. `#jump`, `#force`, `#pray`) , not specific to this PR. The NetHack direction keys also conflict with selection keys for menus of iventory, naming, kill list, etc. I might try to fix them all some day.

fejoa:
> I initially thought on x X, didnt chose cuz of wizard, but.. as "~" is wizard autoplay, "`" (same physical key) will be great for wizard!

Yeah "`" will be good for wizard mode

AquariusPower:
> The NetHack direction keys also conflict with selection keys for menus of iventory, naming, kill list, etc. I might try to fix them all some day.

I added this: `game::ValidateCommandKeys()`
it catch the 'b' !
but I guess what you are talking is other kind of conflicts right?
May it can be improved I guess.

iology:
This is not ready. After I dropped my weapon and polymorphed it, openning the manager crashed the game.

fejoa:
This is awesome. 

'''
Tests required:

    will cause problems if item was offered (destroyed) to a god?
    will cause problems after throwing or dropping the item?
    will cause problems if item breaks after being assigned?
'''

Are these problems fixed? If not, are they anticipated to cause crash?

AquariusPower:
throwing/dropping there is no problem as I remember, neither offer to god, and as breaking or completely rusting will "send it to hell" and be deleted from memory, I think those are ok too. The whole point was to make sure they exist thru the consistency lists.

AquariusPower:
@iology 
https://github.com/Attnam/ivan/pull/401#issuecomment-396474730

I was thinking,

the commands could be set thru a define that also catches the function name like `EquipmentScreen` and transform it into a string.

with that, using the console commands https://github.com/Attnam/ivan/pull/458
there could have a command to bind keys like
`bind E EquipmentScreen`
`listbinds`
`unbindkeys`
and a config file that would be loaded once when the 1st game is loaded in that session (static bool)

I never used other key-sets than the default first one.
so I am not sure this customizability could help?

iology:
> I never used other key-sets than the default first one.
> so I am not sure this customizability could help?

Yeah, I don't feel the urge to do this yet. GUI is always better than just text — select from the list of actions/commands and capture the keypress — then save the config as text.

AquariusPower:
@red-kangaroo 

> This sounds nice. BTW, thinking about weapons and QOL features, is it possible to have autopick for weapons you have thrown, so that you don't need to pick them one at a time?

https://github.com/Attnam/ivan/pull/460/commits/698ea97996893a1081bf272d4240039a113842fa

:)

Issue #399, opened 2018-06-11
Fantasy names: allow <> and | chars in the name pattern


fejoa:

iology:
It is blocked by `iosystem::IsAcceptableForStringQuestion()` in FeLib/Source/feio.cpp

I don't understand the logic there, maybe they are some harmful characters somewhere else? I tried viewing the history of the CVS code but couldn't find any comment about that.

fejoa:
Some trivia: you can bypass this by setting the `FantasyNamePattern` for example to `"!<(ham)|(ster)>i !sV"` up in the `ivan.cfg` file

AquariusPower:
> It is blocked by iosystem::IsAcceptableForStringQuestion() in FeLib/Source/feio.cpp

thx on the tip!

PR #404, opened 2018-06-12
Fix fantasy name pattern input

AquariusPower:
https://github.com/Attnam/ivan/issues/399


AquariusPower:
ugh... felib shouldnt know about anything outside it...
the parameter named fantasyname makes no sense...

iology:
Yeah, ugly code. I suspect the old mysterious logic is related to the bugs before I change IsAcceptableForStringQuestion(**char*) to IsAcceptableForStringQuestion(**int**).

iology:
Not related to missing font glyphs too. Checked all call sites but found nothing special. I think that restriction can be safely removed.

iology:
Oh, I haven't checked the config file: `inputfile::ReadWord()` in FeLib/Source/save.cpp

Definitely have problem with `stringoption::SaveValue()`.

AquariusPower:
'<<' and '|' are present in defines.dat, could be related I guess.

`stringoption::SaveValue`
`SaveFile << '\"' << Value.CStr() << '\"';`

mmm... may be we could just remove the restriction and make tests see if anything breaks.

if you find a better way, feel free to propose another PR :)

I usually put these things as optional (like that behavior simple modifier) to skip the research and tests step btw :)

AquariusPower:
closed in favor of #405, 
I can re-open if needed :)

PR #405, opened 2018-06-13
(ready): Allow "<" ">" "|" in fantasy name pattern

iology:
ping #399


AquariusPower:
interesting documentation.
I use a lot the "valid chars" 0x20 to 0x7E

iology:
Doc/Obsolete/Bugs.txt
178://(hs) Choosing a name including one of characters '?', '*', '|', '<' or '>' and saving will crash (at least in Windows).

Do these keys still crash the game now?

iology:
The special symbols are not allowed in filename. The save files should use safer names.

iology:
This PR is ready now.

Issue #237, opened 2017-03-01
Breaking poly wand explosion crash


Coolthulhu:
Not sure about the exact mechanics, but it reproduces semi-reliably even without trying (ie. in regular game):
* Grab wands. Be sure to include polymorph and at least one damage wand (lightning, striking, fireballs). The more explosives the more likely it will trigger.
* Get exploded (activate a backpack)
* (Hopefully) You'll turn into something and then the game will crash

The last message hints at breaking wands, but not unambiguously: there is always a "wand breaks" message, but it isn't always the last message on the list.

Further possible hints:
* I was unable to reproduce this with wands on the floor, despite repeated attempts
* I was unable to reproduce this while already polymorphed

May take a few tries because it requires the wands to break - possibly in specific order.

fejoa:
Awesome bug finding! Thanks for the hints!

fejoa:
Is this possibly related to #107?

@emlai as an aside, what debug tools are you using?

emlai:
@fejoa lldb.

> Is this possibly related to #107?

Quite likely.

emlai:
Ok here's what's going on:
1. The explosion breaks the wand of polymorph, which causes the player to be polymorphed.
2. As a consequence of this, the old form of the player gets removed from the game and replaced with a new form.
3. Because the old form is removed from the game, its `SquareUnder[0]` is set to null in [char.cpp:7605](https://github.com/Attnam/ivan/blob/1d59ce1a787a713c196e58f5e640fdfeb2ce347a/Main/Source/char.cpp#L7605).
4. Finally the execution returns back to `character::GetHitByExplosion`, where `this` refers to the _old_ form, whose `SquareUnder[0]` is null.
5. Then [`GetArea()`](https://github.com/Attnam/ivan/blob/1d59ce1a787a713c196e58f5e640fdfeb2ce347a/Main/Source/char.cpp#L7171) gets called, which accesses `SquareUnder[0]`, and boom.

So this did originate in [this commit](https://github.com/Attnam/ivan/commit/0a73d297788cb0436516f5e1aab0c406284c7845#diff-4cadc0208ffad4bddcb5b649814f7d9bR7002) as found by @fejoa in #107.

fejoa:
What fun! It does require a special set of circumstances to cause the bug. I wonder how best to fix it? We could revert the change, or change the change, or cause polymorphed characters to immediately initialize their square under?

emlai:
[...]

Issue #107, opened 2016-03-27
Null pointer crash in character::GetHitByExplosion


emlai:
`SquareUnder[0]` of the player character is null with the following stack trace:

```
* thread #1: tid = 0x45c40, 0x00000001000373e2 ivan`square::GetArea(this=0x0000000000000000) const + 66 at square.h:39, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x20)
  * frame #0: 0x00000001000373e2 ivan`square::GetArea(this=0x0000000000000000) const + 66 at square.h:39
    frame #1: 0x000000010003264f ivan`character::GetArea(this=0x0000617000021480) const + 31 at char.h:703
    frame #2: 0x000000010055a237 ivan`character::GetHitByExplosion(this=0x0000617000021480, Explosion=0x0000606001a4e6c0, Damage=300) + 4903 at char.cpp:7006
    frame #3: 0x0000000100398ed4 ivan`level::Explosion(this=0x0000613000164e80, Terrorist=0x0000617000510580, DeathMsg=0x00007fff5fbfc7e0, Pos=(X = 39, Y = 14), Strength=300, HurtNeutrals=true, FireOnly=false) + 6516 at level.cpp:907
    frame #4: 0x0000000100746c4e ivan`holybanana::ReceiveDamage(this=0x0000610000088540, Damager=0x0000617000510580, Damage=42, Type=32, (null)=8) + 2238 at miscitem.cpp:2359
    frame #5: 0x00000001001a8fbc ivan`stack::ReceiveDamage(this=0x00006060001eb780, Damager=0x0000617000510580, Damage=42, Type=32, Direction=8) + 2012 at stack.cpp:321
    frame #6: 0x00000001002dc4e2 ivan`humanoid::ReceiveDamage(this=0x0000617000021480, Damager=0x0000617000510580, Damage=42, Type=32, TargetFlags=255, Direction=8, Divide=false, PenetrateArmor=false, Critical=false, ShowMsg=true) + 4402 at human.cpp:1395
    frame #7: 0x00000001003df8cb ivan`lsquare::Strike(this=0x000061400025d440, Beam=0x00007fff5fbfd860) + 1771 at lsquare.cpp:1510
    frame #8: 0x00000001002fd61e ivan`darkmage::GetAICommand(this=0x0000617000510580) + 20318 at human.cpp:3457
[...]

PR #412, opened 2018-06-16
(ready) Fixing many "on death" crashes

AquariusPower:
related (but not equal): 
https://github.com/Attnam/ivan/pull/240 
https://github.com/Attnam/ivan/issues/107 
https://github.com/Attnam/ivan/issues/237

to simulate it:
have at least one polymorph wand on inventory
drop a landmine
apply a dry backpack of gunpowder

you may have to polymorth in a non humanoid, I tested with "snake" (as I was testing the other PR too).
that explosion may not be strong enough to destroy bodyparts so they may just be severed,
and while polymorphed the placing of the bodyparts on floor will crash as the player body was removed,
it will try the polymorph backup and finally just store them on inventory as failsafe gum solution...

There are other usage crashes, when using look mode after death, etc, but all related to having polymorphed or having the player removed, most crashing basically by Player->GetPos() failing as there is no "square under".


iology:
Oh, great. Did the game crash for the death of monsters as well?

AquariusPower:
I havent tried to xplode a mob,
I rarely use landmines/backpacks, I keep them for later or sell them,
but from what I read on the code, that most probably happens with mobs too so I guess it is fixed,
(unless I missed some detail).

PR #413, opened 2018-06-16
(ready) Mini map notes

AquariusPower:
using the engrave subsystem,
the minimap (immersive mode) will now be able to display notes.

The non immersive (if someone is using), still needs deciding better notes placement.

to create a note, you have to be where the note should be pointing to.

using the position question mode, we can add notes elsewhere.

rotating notes around the map

the basic notes system is ready

PS.: should not use the engrave system? btw, for what is it used at all? once I thought we could "talk to golems" thru it, but never worked, wonder why.. ? xD

PS.2: the engrave system could be also expanded (one day, other issue) to let we set patrolling locations to assign Pets to them.


AquariusPower:
the added mouse support will just check for it's position when in map mode.
the highlight of notes (and target lines) lets the options "delete" and "edit" to work.
the option "edit" will also work w/o highlighting on the current player location.

should not use the mouse? but was so much easier that I preferred it :), keyboard support to navigate notes can be added later.

2018 - Ivan has 1st mouse support actions xD

PR #418, opened 2018-06-19
Limit golem spawns to gloomy caves and tomb of xinroch

### LINKED ISSUE #417: Prevent golems from appearing in Underwater Tunnel
https://attnam.com/posts/29134

fejoa:
Fixes #417

### DIFF
--- a/Script/char.dat
+++ b/Script/char.dat
@@ -2808,6 +2808,7 @@ golem
   IsEnormous = true;
   CanChoke = false;
   UndeadVersions = false;
+  AllowedDungeons = { 2, XINROCH_TOMB, ELPURI_CAVE; }
 
   Config VALPURIUM;
   {


iology:
Too strict, maybe? Earlier, before that surprise was introduced, I could see dangerous monsters if I stayed in the UT too long. Was this feature also removed in a previous commit?

red-kangaroo:
I like that. Golems should not spawn that early.

@iology You will still see dangerous monsters other than golems.

andrewtweber:
Just to clarify, will they still show up in UT4 (the crystal level?)

red-kangaroo:
@andrewtweber The golems in UT4 and the one guarding the secret chest are pre-defined and will be generated even if we forbid golems from spawning in UT.

Issue #374, opened 2018-05-30
festring operator= sponginess


AquariusPower:
EDIT: experimental branch https://github.com/AquariusPower/ivan/tree/FestringCopiesOnly

1st evidence in old save list code (what bug?)
```
      /* Buffer = ep->d_name; Doesn't work because of a festring bug */
      Buffer.Empty();
      Buffer << ep->d_name;

```

to test what happens after reading operator= code behavior:
```
PlayerName="test";
festring name = PlayerName;
name[0]='b';
// PlayerName becomes "best" ...
// even if you use this:
festring name = PlayerName.CStr();

```

expected behavior would be to copy the other festring `char*` value I guess...

a lot of places seem to use it in a non harmful way.

should we change it and check  every spot on the code using operator= ? may be.
should we extend std::string into a new festring? may be.
should we keep it as it is? may be :), learning curve and so on...

or did I miss something that makes the current operator= behavior be important?

now get a new developer, coding a huge lot of stuff, and 3 months later discover about it? scary...

EDIT:
may be one day: remove the 2 operator= 
to substitute it where essentially required create methods .AbsorbHook(char*) and .AbsorbHook(festring&)
what will break everywhere using operator= and be easy to change.
unless that's not a good idea :)

EDIT: and a new method something like: SetInternalDataAsCopyOf(char* or festring&)

iology:
This is suspicious: https://github.com/Attnam/ivan/blob/b973bc99ab6a6a9bc3144fd29f67d286c0eb9caa/FeLib/Source/festring.cpp#L84

I guess that line should be `if((Data == StrPtr) && (OwnsData == Str.OwnsData))`?

iology:
Line 72 and 80 are also problematic: https://github.com/Attnam/ivan/blob/b973bc99ab6a6a9bc3144fd29f67d286c0eb9caa/FeLib/Source/festring.cpp#L63-L91

The current `operator=` does not copy strings when the string data points to a `const char*` created at compile time, or when the string is already owned by another festring. It uses a reference counter to track if the data is being shared by others.

iology:
To be really serious, we need to write a lot of tests for festring.cpp but I feel so tired. :P

iology:
Or only refine those methods which can alter the festring, to make operator= "lazy" (copy-on-write). I haven't checked how frequent operator= is used.

AquariusPower:
> write a lot of tests for festring.cpp

to precisely catch the bug would be good indeed

[...]

PR #430, opened 2018-06-21
(to test) FeString copies only

AquariusPower:
Why? https://github.com/Attnam/ivan/issues/374

The best think would be to spend a lot of time manually scanning the code for problematic places that could use festring in "absorption mode", therefore depending on that behavior to work correctly, to not crash.

I decided to create this PR so it can be more easily tested by everyone.
I've been running all my tests and everything else from all other PRs/branches using this code and had not other problems than the one fixed already (that actually would probably only happen w/o this patch).


iology:
I'm trying to improve the original behavior (to save memory) and to fix that CStr() problem.

AquariusPower:
closed in favor of https://github.com/Attnam/ivan/pull/433
the validator may still be useful, it keeps checking every assigned festring value for invalid characters, I drop here the code:
```
/**
 * code to track new festring assignment (that makes a copy of the string instead of absorbing it) possible bugs
 */
void bugTrackInvalidChars(cchar* c,int size){
  if(!c || !size)return;

  //symbols and characters that are not "usually" expected
  for(int i=0;i<size;i++){
    static char C;C=c[i];
    if(
        (C < 0x20 || C > 0x7E)
        && C!='\r' && C!='\n' && C!='\t'
        && C!=0
    ){
      ABORT("string contains \"probably invalid\" chars, possibly a bug: \"%s\", [%d], 0x%X /0%o %d '%c'",c, i, C,C,C,C); //actual char '?' must be last as may break the string!
    }
  }
}
```

Issue #283, opened 2017-11-07
Direction keys appear to press twice


andrewtweber:
For certain actions (so far I've discovered it with chatting and applying a stethoscope), after you press the direction key, it appears to take the input twice.

So for example when I tried to apply a stethoscope to myself using `numpad 5` it would flash the "information about you" and then immediately close it. (If I used the `.` key instead it worked)

When I tried to chat with the blacksmith in Attnam, I had to specify direction because another NPC was next to me. After I specified the direction, the blacksmith asked what I wanted to have repaired but that dialog immediately closed before I could choose an item.

fejoa:
Arrgh! That's related to the numpad fix. Probably needs a check for whether ```'g'o``` is active

andrewtweber:
@fejoa I actually don't think it's directly related. I tried reverting your change to `Main/Source/char.cpp`

https://github.com/Attnam/ivan/pull/276/files#diff-4cadc0208ffad4bddcb5b649814f7d9b

So that `go` and `dig` are broken again. But applying a stethoscope to myself still detects the `numpad 5` key twice.

andrewtweber:
Fun update: this only affects Mac (/Linux) not Windows

andrewtweber:
It's working!

PR #433, opened 2018-06-21
(ready): Improvements on festring

iology:
ping #374
alternative to #430

haven't checked the syncing problem, e.g. two festrings use the same underlying data and one of them modifies the data and hopes the other one share the modification


AquariusPower:
https://github.com/Attnam/ivan/pull/430#issuecomment-399027783, "to save memory" I guess the only safe way (keeping absorption) would be the absorbed string to remain 'const' in some way right? but I am not so used to cpp to actually know how it should be. 
The code looks interesting, I would like to have understood more of it but as I learned a lot coding Ivan, it just means I am still learning :)

iology:
> #430 (comment), "to save memory" I guess the only safe way (keeping absorption) would be the absorbed string to remain 'const' in some way right?

We have to introduce a new data type. Now `OwnsData` is only set to `false` when that special data (festring::staticstring) is used or when `Data == 0`. A small problem is still left in the old code: strlen does not check NULL (0) so festring may OwnsData even if `Data == 0`.

`consexpr festring::staticstring()` ensures `strlen()` is called at compile time. (C++11 feature)

AquariusPower:
<details><summary>I was reading slowly...</summary>

> OwnsData is only set to false when that special data (festring::staticstring)

so absorption will still be available in a safe way if anyone wants to use it? nice!

> strlen does not check NULL (0)

may be we could create festrlen()? I couldnt find the code to drop a scratch here, I think I have only the headers installed..
</details>
.

> make sure str is a string literal "something"

that looks safe!

PS.: if I am not wrong, internal festring data may have no `0` terminator as I saw some weird outputs in the debug log (that didnt consider the festring size as limit), but I didnt made more tests to make it sure. I guess it is not a problem tho, just another unusual thing worth mentioning :)

AquariusPower:
> the content will always be **copy/read only**.

cool!

> treat NULL pointers as normal char*

I couldnt visualize it (as in an usage example) but it sounds interesting :)
would be like festring("") == festring(NULL) == festring(0) ?

iology:
I'm stuck at `rawbitmap::PrintfUnshaded()`. 🤕 Everything looks so normal when I comment out that function.

AquariusPower:
> So using a null pointer has undefined behavior. Or we can abort the game when it is met.

it is quite important to give users a nice message so they have a better report to help us back. ABORT() will be the best thing for sure. Undefined behavior may even lead to corrupted savegames I guess, and it would be the worst scenario :>

iology:
Now NULL is checked when modifying the `Data`. For functions like `Find(cchar*, sizetype N)`, treating NULL as "" is weird when N is not zero — that may be a problem if nobody knows why.

AquariusPower:
I merged all my PRs and this one here: https://github.com/AquariusPower/ivan/tree/THE_MERGED_BRANCHES_I_AM_PLAYING%2BStrongFeString
I let wiz autoplay on it for some hours w/o problems :)
locally I also  added that check for invalid chars during festring::operator= and it didnt ABORT() either.

AquariusPower:
I found something interesting, I didnt try to track too much tho.
If I enable wizard mode and try to read the scroll of wishing, it instantly asks me if  I want to cancel reading and I can only press 'y'.
Does it happen in your branch alone?

iology:
@AquariusPower No. Can #283 be related? Did you change some parameters about the detection interval of key press?

AquariusPower:
thx for digging that out, 
the problem description there is quite annoying,
I will take a look on it!

I did a "wait key up" to let it work correctly on linux, and is compile optional.
also I added a timeout to let mouse auto-select felist entries on hover.

but... the problem I described is not happening anymore... and nothing was changed in-between...
I think if I just had restarted the game it would have vanished, but instead I changed branches and recompiled...
So it may be some SDL vs Linux input (random?) issue, or some other random linux issue that happened only that time (I cant recreate the problem :blush: ).

fejoa:
This is the next thing to be merged. Please resolve conflicts (last time I did this myself, it resulted in poor personal performance on my behalf).

AquariusPower:
I think there is nothing complex (about conflicts) in feio.cpp and game.cpp, 
were adjustments based in the fact that we can now trust on the fixed festring class will work as expected and consistently
so we dont need to do things like this anymore:
`festring fs; fs<<"test";`
we can just 
`festring fs="test";`
:)

iology:
Hooray~ Time to fix other intricate bugs.

fejoa:
Great work folks!

PR #434, opened 2018-06-23
(ready) Lower Resolution to better fit scaled window

AquariusPower:
lowest resolution possible now is 640x480 allowing window scale x2 to fit better on FullHD monitors

I am using this, so it fits well in 1920x1080, using window scale x2 and the smooth scaling option. Also not full screen to I can keep a watch on my system load at the systray! :)
```
WindowWidth = 960;
WindowHeight = 510;
GraphicsScale = 2;
FullScreenMode = 0;
ScalingQuality = 1;
```

The main menu background also gets centered for lower res to look good.

Bonus: string config options now have the current value pre-filled to help just changing them

Bonus: wizard autoplay fastest mode is now ultrafast but stoppable, it will play as fast as your machine can handle now (just disable xBRZ)! Obs.: added gum solution to weird slowed wizard autoplay w/o xBRZ enabled, still no idea why that happens :/

PS.: ugh.. it is getting a bit confuse now to maintain so many branches :>, I will hold a bit before it gets too complicated xD


FeLib/Source/feio.cpp
```diff
@@ -163,15 +163,28 @@ int iosystem::Menu(cbitmap* BackGround, v2 Pos,
...
     if( (RES.X!=BackGround->GetSize().X) || (RES.Y!=BackGround->GetSize().Y) ){
+      blitdata B = DEFAULT_BLITDATA;
+      B.Bitmap = &Buffer;
+
+      B.Src.X = (RES.X - BackGround->GetSize().X)/2;
+      if(B.Src.X>0)B.Src.X=0;
+      if(B.Src.X<0)B.Src.X*=-1;
+      B.Src.Y = (RES.Y - BackGround->GetSize().Y)/2;
+      if(B.Src.Y>0)B.Src.Y=0;
+      if(B.Src.Y<0)B.Src.Y*=-1;
```
iology: What do Src.X and Src.Y mean?

AquariusPower: blitdata should have better named var members:
```
  bitmap* Bitmap; //target bitmap to draw in
  v2 Src; //pos at source bitmap to copy from
  v2 Dest; //pos at target bitmap to draw in
  v2 Border; //size to copy from at src bitmap
```
we could just add these comments there too :)

Main/Source/iconf.cpp
```diff
@@ -666,7 +666,7 @@ void ivanconfig::ShowItemsAtPlayerSquareChanger(cycleoption* O, long What)
 
 void ivanconfig::WindowWidthChanger(numberoption* O, long What)
 {
-  if(What < 800) What = 800;
+  if(What < 480) What = 480;
```
iology: You swap the width and height...

FeLib/Source/feio.cpp
```diff
@@ -163,15 +163,28 @@ int iosystem::Menu(cbitmap* BackGround, v2 Pos,
...
+
+      B.Dest.X = (RES.X - BackGround->GetSize().X)/2;
+      if(B.Dest.X<0)B.Dest.X=0;
+      B.Dest.Y = (RES.Y - BackGround->GetSize().Y)/2;
+      if(B.Dest.Y<0)B.Dest.Y=0;
+
+      B.Border = BackGround->GetSize() - v2();
```
iology: Just recompiled the game. So finally it clips the background image by starting at the top left corner, or centers the image. Can you scale it?

AquariusPower: you mean shrink?
I think there is no code in ivan capable of shrinking, not sure tho.
But I think SDL has it, but the image have to be converted to their format first, and I think it is 24bit color. The xBRZ lib does such conversion and a function there could be used to prepare the background.
Another option is just to shrink the menu image using Gimp for ex. :)

Main/Source/iconf.cpp
AquariusPower: ugh.. I missed this comment.. I guess github could highlight the news on the page we are reading? may be suggest at github forums :)

PR #440, opened 2018-07-02
(ready) Fix to allow new game with a different player name, with equal name and to load AutoSave(s) (and just after it, save) properly again

AquariusPower:
also:

improved the filename to show readable game creation date time

added a fix to allow **starting a new game** with the same player name in a session that another character with that name was just played (so that a new game will be created now instead of loading the last one played).

added checks and fixes to the savegame basename for consistency to not interfere with the complicated ".AutoSave" suffix :)

Obs.: the PR error is about travis-ci and opengpg download etc, not the code (on my PC at least)

PS.: I expect the title is not too confuse xD


Main/Source/game.cpp
```diff
@@ -3270,50 +3277,93 @@ int game::Load(cfestring& saveName)
...
-  festring SaveName = GetSaveDir();
+  for(festring::sizetype i = 0; i < fs.GetSize(); ++i)
+  {
+    if(fs[i]>='A' && fs[i]<='Z')continue;
+    if(fs[i]>='a' && fs[i]<='z')continue;
+    if(fs[i]>='0' && fs[i]<='9')continue;
+
+    fs[i] = '_';
+  }
+}
```
iology: Really hope #433 will be reviewed and merged soon. `fs[i] = X` is still harmful.

Main/Source/game.cpp
```diff
@@ -3270,50 +3277,93 @@ int game::Load(cfestring& saveName)
...
+
+bool chkAutoSaveSuffix(festring& fs,bool bAlsoFixIt=false){DBG1(fs.CStr());
+  std::string strChk;
+  strChk = fs.CStr();
+  int i = strChk.find(AUTOSAVE_SUFFIX);
```
iology: You can also use festring::FindLast()

Main/Source/game.cpp
```diff
@@ -3270,50 +3277,93 @@ int game::Load(cfestring& saveName)
...
 #endif
 
-  return SaveName;
+  if(!bLoadingFromAnAutosave){ //very specific use case
+    if(chkAutoSaveSuffix(PathAndBaseSaveName)){
+      /**
+       * this ABORT is important to prevent the troubling autosave suffix duplicity,
+       * it's consistency is kept using only: GetAutoSaveFileName()
+       */
+      ABORT("The base savegame filename '%s' must not contain '%s'",PathAndBaseSaveName.CStr(),AUTOSAVE_SUFFIX);
```
iology: "must not *end with*"?

Main/Source/game.cpp
AquariusPower: I will take a look on it thx

Main/Source/game.cpp
AquariusPower: we actually need a findFirst thing as we can have "PlayerName.AutoSave.AutoSave" use case,
I should have implemented it into festring but I found festring code quite difficult to follow when I thought on doing that in the past, so better not change what I dont understand well :)

Main/Source/game.cpp
AquariusPower: some funny user can create his player name like ".AutoSave funny guy", better avoid it if possible :>,
a better way would be to have more checks, so or it or more constraints :)

Main/Source/game.cpp
iology: Without my PR, you can create a local temporary festring, manipulate the temp string and then `fs.empty(); fs << temp;` If my PR get merged first, then you don't have to change anything.

`festring::sizetype` is unrelated to the size of `char`. "character/glyph" and `char` are different things. It can already work for all language if the encoding of input text does not conflict with this single byte manipulation of ASCII text.

Main/Source/game.cpp
iology: Then `festring::Find()` and `SEARCH_N_REPLACE` is enough. Also need to check if the final string is empty.

Mixing std::string and festring is a bit funny. :P

AquariusPower: I was thinking, SearchAndReplace() will work if nothing unexpected happens.
"player_name.AutoSave.hello" will provide unpredictable result
I dont know.. I like to do what is safest (considering funny users typing funny things just to see what happens :)) even if beyond the scope of the function name "chkAutoSaveSuffix()".
Feel free to restrict the scope, the beyond things may never happen anyway :).

iology: Need a while loop to fix "..AutoSaveAutoSave.AutoSave"? But "." will be replaced with "_" for save files, so such names will only be detected in two situations:

1. Loading a save file from an unknown version of the game. (with the user's own patch?)
2. The save file was manipulated by the player outside the game.

I think there is no need to fix the name because it is very unlikely that the fixed name will match an existing file.

AquariusPower: the fix for unprobable situations, despite useful, is not required.

but it is required on this flow:

```
truth game::Init(cfestring& loadBaseName){
  festring absLoadNameOk;
  if(!loadBaseName.IsEmpty()){
    absLoadNameOk = SaveName(loadBaseName,true); //will prepend the path
...
festring game::SaveName(cfestring& Base,bool bLoadingFromAnAutosave){
...
     chkAutoSaveSuffix(CurrentBaseSaveFileName,true);
```

becase the chosen savegame to be loaded may be an .AutoSave, and it is important to keep the CurrentBaseSaveFileName w/o that suffix

AquariusPower: btw, I opted for using std::string because of it's quite useful `.substr(` method, but indeed I dont expect that after .AutoSave has anything other than another .AutoSave :)
this means, if one day there is something, it will be missing and such code will have to be modified.
but... I suggest we dont put anything after .AutoSave just to not make things more complicated xD

iology: Ah, okay. That function name is misleading… It "adjust"s, not "fix"es the filename. Then you should only erase ".AutoSave" at the end of the string, while still searching without any restriction for the return value `true`/`false`. If ".AutoSave" does not end the string (theoretically), then you will get a false-positive result for ".AutoSave fun" on Line 3357.

AquariusPower: `If ".AutoSave" does not end the string` 
mmm... if .AutoSave is anywhere else than in the end, ABORT(), this may be good! and dont sound that complex

iology: Just realized there are also ".AutoSave.sav", ".AutoSave.sav.bkp" and the like. :P

AquariusPower: and the quite hard to see .AutoSave.sav.tmp (crashing on creating it will keep .AutoSave.sav and the .bkp untouched :))

AquariusPower:
@fejoa 
If I am not wrong, w/o this PR I was unable to start a new game using the random fantasy name (I this it was after loading another, saving to main menu and trying to start a new one)
I fixed the conflicts (as also with all other open PRs, we still have no pending cuz  of conflicts indicator/notification it seems but I asked for something at git forums https://github.community/t5/How-to-use-Git-and-GitHub/pull-request-requiring-conflict-resolving-should-have-a-nice/td-p/7800 :)).
So this one at least would be good to be merged soon :)

fejoa:
I noticed this bug before, thanks for the fix @AquariusPower :)

PR #443, opened 2018-07-08
(ready) Enhanced lights

AquariusPower:
distant lights will be visible and also their nearby squares.
can be disabled.

Bonuses (too many active branches, I will put all new and working simple stuff on this PR too): 
- fixed wiz autoplay to actually play in background
- changed default safeSaveGame=true (also in all other open PRs)
- fixed old bug when chameleon with polymorph locked would crash (should a chameleon be immune to polym. lock instead?)
- opt to limit m.mushrooms AI https://github.com/Attnam/ivan/issues/446
- easier kicking (press 'k' (or your setup kick key) again to chose the previous direction chosen
- easier throwing: will remember last selected index. after hit enter to chose item to throw, just hit enter again to chose the previous direction chosen

PS.: to test all my PRs together, use this branch:
https://github.com/AquariusPower/ivan/tree/THE_MERGED_BRANCHES_I_AM_PLAYING
or
https://github.com/AquariusPower/ivan/tree/THE_MERGED_BRANCHES_I_AM_PLAYING%2BStrongFeString
(the last one updated)


Main/Source/level.cpp
```diff
@@ -2946,6 +2948,131 @@ void sunbeamcontroller::ProcessStack()
...
+          Square->Reveal(LOSTick,false);
+          ++tot;
+        }
+      }
+    }
+  }
+
+  return tot;
+}
+//int level::RevealDistantLightsToPlayer()
```
AlexMooney: Avoid committing commented out code.  Presumably this stuff was checked in at some point.  If you need it again, you can pull it out of the history.

AquariusPower: actually this is a new feature I created and as I remember I did not commit the previous working methodology :)
I left it there like that as could help on improving the main code as it still had some interesting ideas, but ended not being practical for now.
mainly this:
```
//        /**
//         * the farer, less range around the emmiter can be seen
//        long lDelta = lMaxDist - iMinDist;
//        float fPerc = (iDist - iMinDist)/(double)lDelta;
//        int iLightBorder = 0xFF*(0.75 + 0.25*fPerc); //reference LIGHT_BORDER is 0xFF*0.5, but here is for far things
//         */
```
despite it would be cool to have it working, it got too complicated to implement so I give up on it, but someone may figure out later how to. I kept like that because it was compiling.
if you would like to clean it up, just try to keep these lines as they are still an interesting idea :)

AquariusPower: I moved that part as a TODO comment :)

PR #452, opened 2018-07-20
Added graphic image for vials

Stihdjia:
Now I'm hooked on messing with this stuff, I hope this change isn't too trivial for a pull request. Vials were invisible as they were using an empty space in the graphics file, I made one and also fixed the appearance of the cork in bottles. Apparently bottles have had corks all along but they were drawn with colors that weren't properly displayed in game.


Stihdjia:
Derp it was already fixed... Corks are weird still though

AquariusPower:
feel free to try to improve it,
I think vials could look better,
I copied the bottle graphics and just made it thin xD

Stihdjia:
Lol that's exactly what I did too. Our versions look very similar. Do you
see how the cork/stopper on bottles and vials aren't showing correctly
though? I'm not sure why, but only the darkest pixels are working in game.
I never realized they were supposed to be there.

AquariusPower:
there is 2 kind of colors
col16 the actual graphics
and 
col24 all the luminance effecs
so if I am not wrong, modifying a bit the colors would make it fit on the limited range col16 provides compared to col24, but I dont understand that much about it...

PR #463, opened 2018-08-01
(wait) if consumed starts spoiling, a question will be asked before continuing.

AquariusPower:
I think getting a disease thru just spoiled food while consuming it is part of the challenge,
but when we play very carefully and still get infected that is just annoying,

**may be this should be optional?**

what do you think?

EDIT: finally tested, bear, wait 275 turns, a bit hard to test :>


iology:
So the reason for the TODOs is why I sometimes rest in pain from acute poisoning or mustard gas resolving? Should modify Main/Include/actions.h and record some old states.

AquariusPower:
@iology 
when consuming a corpse that is still edible (no flies), it may start spoiling while you are still consuming what you thought/expected should be edible while you eat it.
With this feature, if it starts spoiling while you are still eating what you thought should remain edible, you will get a message to stop eating it :)

I think one of the problems caused by spoiled food may be poisoning, but never mustard gas or am I wrong?

>Should modify Main/Include/actions.h and record some old states.

When you begin one vanilla action, it is 'Handle()'d. When you stop it, it will always be deleted from memory thru 'Terminate()'.
The new crafting action for some reason is not being saved, but as long you dont exit the game, it can be suspended and continued later.
Unless you mean other kind of 'old states'? :)

iology:
Oh, I mean #474. You can add a JustStarted flag for the first todo.

AquariusPower:
oh thx, 
instead of re-using ActivateInDNDMode() you mean right?
btw, b4 eating, I pay close attention to check for flies, I use the zommed list item option to help on it.
but my monitor is a  bit bad, if my head is not aligned to it's center, the image will be a bit discolored preventing seeing flies...
so I ended lefting it unperfect like that as it would always warn in case we try to eat spoiled food from the beggining,
but it could be perfected indeed with "thing thing is spoiled, continue?" and it would be much better, and that flag (or 2 flags) would help a lot :)

AquariusPower:
> the consequence could be surprising

Im guessing not good ones...

Btw, feel free to patch it (a PR for this PR) if you prefer, I am working on the craft code again and on the Ctrl+ and function keys F1-F12 :), the needed detailed help for config options may be not that far now xD

iology:
There are only two flags for `action::Flags`: TERMINATING and IN_DND_MODE, and they are irrelevant to each other. I don't understand why they need to be save and reload for the game. Any idea?

AquariusPower:
by all I have tested, they dont even seen to be saved at all...
when suspending crafting it will never load it back,
I expect I am doing something wrong otherwise saving current action is a missing feature :>

in_dnd_mode to prevent asking again if user decided to continue
terminating to prevent (probably) a second call from another origin (despite same thread) asking to terminate again what could cause some weirdness w/o it

the "need to be saved" part of the question is now unclear, until we be sure it actually loads anything about the suspended action xD

iology:
I can push to your branch directly?

AquariusPower:
I think you can merge what I did into a new branch or yours and feel free to create a new PR.
or in an existing branch, up to you.
or if you already entirely coded it anew, just link to this PR as alternative and I close this one.
once I did a PR to a branch of another user that was already a PR and it just required it be accepted by the other user, what is another option.
I can't see a way to allow you to do this here directly tho, but as you have the "Member" tag I guess you can! not sure tho..
:)

AquariusPower:
closed in favor of https://github.com/Attnam/ivan/pull/477

PR #464, opened 2018-08-01
(ready) auto disable xBRZ during actions

AquariusPower:
may be many ppl avoids using it because actions are quite slowed because of xBRZ,
so it will be auto disabled and restored just after them,
like drinking, eating, digging etc

bonus: https://github.com/Attnam/ivan/issues/465


fejoa:
@AquariusPower can you please the resolve conflicts for this PR?

AquariusPower:
I was w/o internet for some days xD

PR #473, opened 2018-08-03
(wait) FeList filter

AquariusPower:
generic, will work on any FeList,

may still have some glitches,
try dropping things after filtering 
and at the user config list too
etc

not perfect, but a lot useful (if no more bugs are found) :)

Ctrl+F


AquariusPower:
@iology 
was you that implemented this?
```
     case SDLK_HOME:
     case SDLK_KP_7:
      KeyPressed = KEY_HOME + 0xE000;
      break;
```

I much rather prefer returning Ctrl+ KEY instead of the hardcoded function calls at whandler.cpp I created both here and to access the developer commands console on the other PR, but I am not sure how to implement it properly :/

iology:
> was you that implemented this?

No, it is from the original IVAN source.

0xE000 was added to unify the normal keys with the keys of the number keypad. See `Key - 0xE000` in  `globalwindowhandler::GetKey()`.

AquariusPower:
oh thx!
well, I think the flow should be different anyway and will throw Ctrl+KEYS and F1-F12 on a new special events class.

AquariusPower:
unifying special keys at https://github.com/Attnam/ivan/pull/458

PR #474, opened 2018-08-03
Warn about severe injury from mustard gas while resting

iology:
I hate mustard gas. :-/

### DIFF
--- a/Main/Source/char.cpp
+++ b/Main/Source/char.cpp
@@ -11982,10 +11982,16 @@ void character::ReceiveMustardGasLiquid(int BodyPartIndex, long Modifier)
       {
         ADD_MESSAGE("You're about to die from mustard gas.");
 
-        if(game::TruthQuestion(CONST_S("Try to find a healer? [Y/n]"), YES))
+        if(game::TruthQuestion(CONST_S("Mustard gas is dissolving your flesh."
+                                       " Continue resting? [y/N]"),
+                               NO))
+        {
           Action->Terminate(false);
+        }
         else
+        {
           Action->ActivateInDNDMode();
+        }
       }
     }
   }
--- a/Main/Source/char.cpp
+++ b/Main/Source/char.cpp
@@ -11972,6 +11972,18 @@ void character::ReceiveMustardGasLiquid(int BodyPartIndex, long Modifier)
                             BodyPartIndex, YOURSELF, false, false, false);
       CheckDeath(CONST_S("killed by a fatal exposure to mustard gas"));
     }
+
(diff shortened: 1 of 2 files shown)


AquariusPower:
I think many of the improvements we are creating let us play the game a bit faster, 
having to pay less attention to text messages and some other non clear indicators,
as we have sounds for bad diseases and bigger images etc,
I think this message will be quite helpful too! :)

Main/Source/char.cpp
```diff
@@ -11972,6 +11972,22 @@ void character::ReceiveMustardGasLiquid(int BodyPartIndex, long Modifier)
...
+    if(IsPlayer())
+    {
+      action* Action = GetAction();
+
+      if(Action && Action->IsRest() && !Action->InDNDMode()
+         && BodyPartIsVital(BodyPartIndex) && BodyPart->IsBadlyHurt())
+      {
+        ADD_MESSAGE("You're about to die from mustard gas.");
+
+        if(game::TruthQuestion(CONST_S("Try to find a healer? [Y/n]"), YES))
```
AquariusPower: "Try to find a healer?" may be in dungeons we will not have the time to reach one, what about "I should try to heal myself promptly! confirm? [y/n]" or something in that line. Or it could scan for a healer and open the minimap highlighting him/her xD, interesting that I remember a valpurus (?) angel putting back my arms quite often :>

iology: It seems no NPC has the ability to heal you... potions do not seem very helpful... resting in place is too slow. At least you should keep moving and eat something like magical mushrooms. So "Go see if there is a healing potion nearby? [Y/n]" ?

AquariusPower: or "Stop to heal before it is too late? [Y/n]", I think many of these details (that are not the core code logic) end being a matter of preference and are not that important, in a sense that it is not a requirement to merging at all and are extremely easy to be changed later. :)

iology: I feel "Try to find a healer?" is quite good, though actualy they don't provide any cure for mustard gas damage. :p

andrewtweber:
I like the warning but not the text. Maybe something like "Mustard gas is dissolving your flesh. Continue resting? [y/N]"

Of course you would switch the default to NO. This also matches the "Continue resting? [y/N]" that is displayed if you are interrupted by a hostile monster

AquariusPower:
nice, being non-specific as the new course of action, just dismissing the current one, really sounds better :)

iology:
Okay, I've updated the message.

PR #475, opened 2018-08-10
Patch new materials and NEWS update

red-kangaroo:
I tried to address the comments on the new materials.

NEWS file updated with all the changes that were merged but not included in the file. :)


AquariusPower:
thx! I always forget about the NEWS file :>

PR #477, opened 2018-08-12
Warn about spoiling food when consuming them

iology:
Also make ActivateInDNDMode() not called by actions themselves. This makes it easier to maintain the logic.

ping #463

### DIFF
--- a/Main/Include/actions.h
+++ b/Main/Include/actions.h
@@ -48,12 +48,14 @@ ACTION(consume, action)
   virtual truth AllowFoodConsumption() const { return false; }
   virtual cchar* GetDescription() const;
   virtual void SetDescription(cfestring&);
-  virtual void SetNibbling(truth What) { nibbling = What; }
-  virtual truth IsNibbling() { return nibbling; }
+  virtual void SetNibbling(truth What) { Nibbling = What; }
+  virtual truth IsNibbling() { return Nibbling; }
  protected:
   festring Description;
   ulong ConsumingID;
-  truth nibbling = false;
+  truth Nibbling = false;
+  truth Gulping = false;
+  truth Spoiling = false;
 };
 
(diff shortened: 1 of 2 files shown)


AquariusPower:
interesting, I googled, DND is "do not disturb" and is quite generic.
may be the variables could be named alike but specifically as:
SpoilingDND
GulpingDND
NibblingDND

iology:
Nibbling is only related to where the food was at the beginning. So there are public functions for other game mechanics.

Spoiling and Gulping are for dangers that arise from the current action.

InDNDMode(), after this patch, is only for external conditions.

The names could be changed to IgnoreSpoiled, IgnoreBloated and what? I can't think of a good name for DNDMode. :p

AquariusPower:
not a requirement at all!
what really matters is the logic that provides us missing/improved functionalities, 
these (questionable) "better" naming are unnecessary.

I keep changing naming of functions and vars (mainly the ones I create) quite often to make them more clear to me, this doesnt mean it will be as clear to other ppl and they may further change'm in the future xD

iology:
I thought about unifying internal and external conditions with `std::map<cchar, bool> flags`.

AquariusPower:
I think the point is how long they will be available/required.
as soon an action ends, if the user opted for DND at that time, makes sense anymore as later he may be in other kinds of danger and need that question again.
unless I am missing something? :)

iology:
Yep, I'm not sure how long the DND mode should last for external threats (mainly Kamikazedwarves and named monsters). Maybe extreme dangerous monsters should always be warned (only when resting in place).

AquariusPower:
I added, for the k.dwarvs, a sfx laugh sound in one of the PRs, but still it should be an interruption (despite I think it already is). Btw I try to always keep important burnable/xplodable things on chests.

PR #478, opened 2018-08-28
(ready) stethoscope bodypart material info

AquariusPower:
otherwise we could only know about it when reading some scrolls like change material, right?


red-kangaroo:
Great!

AquariusPower:
@andrewtweber 

I can do that but it would not modify every other place in the this whole big code :)

**lower line count and some ugly code**

I got used to, very long ago, to format code like that to lower the line count, so I could read more code on the screen w/o having to scroll down/up. And many other coding styles I use with the same purpose (making things more "one lined").

Btw, all debug code isnt intended to really exist, they were used to debug and fix, so the moment it is fixed such is not that useful anymore, as also compiling for release removes it all from the binary. So everywhere that has DBG1() and the like, could safely be removed one day, may be even automatically thru `sed`. So to not compromise the line count, I did ugly things like:

```
if(...){
 doSomething();
}DBG1(...); //ugly
```

And that can be really annoying to who is not used to coding like that! xD

**auto format to help us all :)**

Eclipse IDE has an auto-format feature that will re-arrange all the code on the current file following a predefined format template with a single hot-key stroke.

May be, instead of just fixing that spot, it could be done for the whole code, we just need to grant the template is working precisely well before applying it (or will just have to re-apply later).
Do any of you use Eclipse IDE?

Obs.: I just moved to NetBeans because it is working better about code completion in most cases (not perfect yet tho but seems better to configure), it also is not that good to use as Eclipse (mainly to navigate the code) but is working better in most cases. I still have Ivan on Eclipse tho.

> I thought we had something that would automatically catch that.

may be, there is a command line tool that could auto-format instead of blocking/preventing the compilation? ( what is not happening actually but if implemented like that may put some new devs with cool ideas away from the project :/ )

I mean just that, after you got used to a coding style, having to **manually** fix these less important things everywhere (as the code logic is what really matters, w/o it we have nothing) would be quite annoying, but... if we could **automatically** fix it, then it would be **perfect to everyone**! we just need the  code formatter (eclipse?) and the well configured template :) !

**another PR to format it all**

Do you think, another PR to fully format the whole code could be more interesting? 
may be a command line OS independent tool would be better (as Eclipse is by being in java), so we could auto apply the formatting thru cmake just before compiling?

I wonder who knows about such and could give a hand on preparing the auto-formatter with the best c++ coding style possible for readability/clarity @iology ? :)

**Old code's style needs improvements too**

btw, there are some places in the old code where they also try to lower the line count for ex.:

```
  for(...)
    doSomething();
    doOtherStuff();
    doManyOtherStuff();
```

as you see above, and if I am remembering it well.. it was badly formatted like that.
what I remember well is that it took me about 1 hour to figure out I should add { } (for clarity), sounds stupid I know... :), but I had to slow down to actually see the problem and that was quite annoying :P

```
  for(...)
  {
    doSomething();
  }

  doOtherStuff();
  doManyOtherStuff();
```

now, what I mean is: every place should use code blocks, so we never get confused again and lose time on such obvious things...

I am not sure but I think the code style formatter can add { ... } everywhere that has only a single command for such block mainly for `if` and loops.

andrewtweber:
Good call. Yes I think one big PR to format all of the code consistently would be good, as long as we have checks to make sure that future PRs all conform. Also we'd want to merge as many open PRs as possible before doing this. Maybe after the next big release.

And yes I dislike some of the formatting, I think we should include curly braces for every `for` and `if` even if it's just 1 line within the block.

PR #489, opened 2018-10-04
(ready) minor adjustments, additions and fixes

AquariusPower:
I ended coding'em at crafting branch,
but I thought it is better in a separate one,
was the cause of the "complicated" conflict solving :>

some changes are only being used at crafting branch for now, but could still be used elsewhere.

from commit log:
- GetFullFilename() at save.h
- v2() now grantedly initializes as 0,0
- festring bugTrackInvalidChars() in dbgmsg mode only
- better FPS positioning
- game::IsQuestItem() to avoid coded multiplicity
- new option IsAllWeightIsRelevant() to prevent item stacking with different weights
- new option IsShowVolume() for items in cm3
- item.h AddContainerPostFix() to let know how full a container is
- materials.h GetSpoilCounter() CalcNewSpoilLevel() to avoid coded multiplicity
- traps.h GetRemoveTrapModifier() TryToTearDown() for reusing outside this class
- offer to gods now remember position to help on doing it many times
- wishing now remembers previously typed
- better eclipse syntax highlight helper
- DBGBREAKPOINT


fejoa:
Eeek, sorry for leaving this one so long...

audio/midiplayback.cpp
```diff
@@ -1020,5 +1020,5 @@ void _MIDI_fileclose(void)
 void _MIDI_readbuf(uint8_t* position, uint8_t* buf, uint16_t size)
 {
    fseek(midifile, (long long) position, 0);
-   fread(buf, 1, size, midifile);
+   size_t a = fread(buf, 1, size, midifile);
```
iology: `a` is unused?

iology:
My review just finished. :p Blindly trusting your code about auto-play mode and AI.

I wonder where the new sound files came from?

audio/midiplayback.cpp
fejoa: It would seem so

AquariusPower: there was a cpp flag, that I got from one of the comments here from some issue, that complained about `fread()` unused return value preventing compilation, dont remember which one tho, but being harmless I thought it was a good idea to let it be there.

AquariusPower:
@fejoa np :)

@iology the 3 "laugh*.wav" I created using a yamaha keyboard (I think it is an old one PSRE223), by hitting (as I remember) 2 or 3 keys in sequence at a certain short interval, then I edited it using audacity to tweak a bit for variety. I think it is not a problem as the final thing is a mix/edited result and not a pure sample that could be re-used as a base elsewhere, so in short it was my creation, in a sense that, holding a single key would provide a pure sample, but what I did was a composition (like a music), but these stuff scares me and I may be wrong, feel free to remove it if it scares you too :)
  btw, all other new sounds, I created making sounds on my home, and a few (like the siren voice) was also a composition/modified effect from a (as I remember) voice/chorus midi.
  also, all the other original sounds (sfx from Ivan3D as I remember) have no details about their creation if I am not wrong right? I mean, are these details necessary?

PR #492, opened 2018-10-19
Tweraif Freedom victory

red-kangaroo:
It is now possible to win by bringing freedom to Tweraif, and then making sure it stays free. It is a low-ranking win, but at least your village will be safe.

For spoilers, see "Add Tweraif Freedom victory" commit.

BTW, you can also challenge yourself and try to win while only ever entering NA and UT, and no other dungeon.


Main/Source/human.cpp
```diff
@@ -6570,3 +6623,126 @@ truth imp::SpecialBiteEffect(character* Victim, v2 HitPos, int BodyPartIndex, in
...
+                             "their cloaking spell can be replicated. Attnam tried to take our\n"
+                             "goddess away, but she is still strong in our hearts. I have faith\n"
+                             "she will protect this island from valpurians, just as Cleptia did\n"
+                             "for Mondedr.\"\n"));
+
+    game::TextScreen(CONST_S("\"The prayers are simple, but no god can affect the world uninvited,\n"
+                             "and a miracle of such strength requires more power than any priest\n"
+                             "could channel. We need a conduit, something close to Silva herself.\"\n"
+                             "\n"
+                             "\"We need a splinter of the Holy Mango World-tree.\"\n"
```
fejoa:
```suggestion
                             "\"We need a scion of the Holy Mango World-tree.\"\n"
```

Main/Source/rooms.cpp
```diff
@@ -710,13 +710,42 @@ truth bananadroparea::PickupItem(character* Hungry, item* Item, int)
...
+        return true;
+      }
+    }
+    else
+    {
+      ADD_MESSAGE("You feel that the climate is not quite right for growing mangoes.");
+      return false;
+    }
+  }
+  else
```
fejoa:
```suggestion
  
```

fejoa: Just needs to have a default case. That should stop the compile error in clang.

Issue #493, opened 2018-11-05
Moving Linux state dir to user dir


akien-mga:
I'm working on packaging the current release of IVAN for my Linux distro [Mageia](https://mageia.org).

I've noticed that the buildsystem and `Main/Source/game.cpp` define `GetStateDir()` as pointing to `/var/ivan`, while macOS uses `GetHomeDir()` for that. `GetStateDir()` is then only used to store the `Bones` folder and high score file.

Nowadays using `/var` for game data on Linux is not recommended, only system and server applications should have a need for it. User applications are meant to use `XDG_CONFIG_HOME` (usually `~/.config`) and `XDG_DATA_HOME` (usually `~/.local/share`) for that.

A simple fix for this on Linux would be to drop the `!MAC_APP` code that defines `GetStateDir()` as different from `GetHomeDir()` on non-OSX. And maybe drop `GetStateDir()` to replace it by `GetHomeDir()` directly to simplify the code.

I'd also suggest considering the use of `XDG_DATA_HOME/ivan` on Linux instead of `~/.ivan` (with fallback to its typical value `~/.local/share/ivan` if not defined), but that would break compatibility with existing save games so it might not be wanted. Just know that using `~/.<name>` is now deprecated and that the "good practice" is to use `<name>` folders in `~/.local/share` and/or `~/.config`.

akien-mga:
Well, right after I wrote this I had a quick look at the master branch and I saw #490. This is similar to what I describe, and was done for similar reasons (NixOS packaging), but I would suggest going a step further and not making it an option.

This split between `GetHomeDir()` (which is also used on Linux as `~/.ivan`) and `GetStateDir()` seems unnecessary, and probably the result of incremental changes to this code without re-evaluating the choices made two decades ago when `/var` was indeed a place where apps wrote their state (actually `/var/games` for games).

iology:
I think `~/.local/share` is not a good place unless IVAN is also installed at `~/.local`, and normally save files should not be shared.

We can first detect XDG_CONFIG_HOME on linux, if no then fall back to HOME.

Should also give a notice about this change while running cmake.

/ping #495
/cc @Attnam thoughts?

akien-mga:
> I think ~/.local/share is not a good place unless IVAN is also installed at ~/.local, and normally save files should not be shared.

[...]

PR #490, opened 2018-10-06
USE_HOME_FOR_STATE_DIR option added

freepotion:
`USE_HOME_FOR_STATE_DIR` option was added to CMakeLists.txt.

Fully consistent with my pull request to [nixpkgs upstream](https://github.com/NixOS/nixpkgs/pull/47950).

PR #495, opened 2018-11-05
game::GetHomeDir() and game::GetStateDir() were merged

freepotion:
Original `game::GetStateDir()` was deleted, `game::GetHomeDir()` was renamed to `game::GetStateDir()`, all calls are replaced accordingly.
`-DLOCAL_STATE_DIR="${CMAKE_INSTALL_FULL_LOCALSTATEDIR}/ivan"` definition and `USE_HOME_FOR_STATE_DIR` option have been removed from **CMakeLists.txt** as unnecessary.
@iology New pull request corresponding to your initiative! :alien:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,14 +13,10 @@ set(CMAKE_CXX_FLAGS "-std=c++11 ${CMAKE_CXX_FLAGS}")
 add_definitions(-DIVAN_VERSION="${PROJECT_VERSION}" -DUSE_SDL)
 
 option(BUILD_MAC_APP "Build standalone application for MacOS" OFF)
-option(USE_HOME_FOR_STATE_DIR "Statedir will be /.ivan/ in current user's homedir" OFF)
 
 if(UNIX)
   add_definitions(-DUNIX)
   include(GNUInstallDirs)
-  if(USE_HOME_FOR_STATE_DIR)
-    add_definitions(-DUSE_HOME_FOR_STATE_DIR)
-  endif(USE_HOME_FOR_STATE_DIR)
 
   if(BUILD_MAC_APP)
     install(DIRECTORY Graphics Script Music Sound DESTINATION "ivan")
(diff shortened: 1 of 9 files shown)


iology:
I spotted a bug at Line 337 of Main/Source/message.cpp: the extra `"/"` (midi works well on Windows so it is hard to notice). Line 990 of FeLib/Source/feio.cpp is also unfriendly to Windows. So the "home dir" on Windows should be `"./"` instead of a empty string. But I don't know if `./` will cause problems on some old versions of Windows.

Should add a notice with cmake.

Waiting for others' opinions. :)

fejoa:
>I spotted a bug at Line 337 of Main/Source/message.cpp: the extra "/" (midi works well on Windows so it is hard to notice). Line 990 of FeLib/Source/feio.cpp is also unfriendly to Windows. So the "home dir" on Windows should be "./" instead of a empty string. But I don't know if ./ will cause problems on some old versions of Windows.
>
>Should add a notice with cmake.
>
>Waiting for others' opinions. :)

You must mean me? It should work on win7 and greater. I know there are some who play on XP, I don't know if it will break there? I don't have an XP or 9x/DOS to test it out on. Perhaps the best/laziest thing to is to break it first, then fix later.

I think the best thing to do for windows as far as folder structure goes, is just to make it a 'portable' version as much as possible. As in, all the data is local to the directory. Imagine keeping a version on a memory stick, or dropbox, that sort of thing.

freepotion:
> I think the best thing to do for windows as far as folder structure goes, is just to make it a 'portable' version as much as possible. As in, all the data is local to the directory. Imagine keeping a version on a memory stick, or dropbox, that sort of thing.

It seems to work that way now. But I'm not entirely sure. :neutral_face:

akien-mga:
It would have been less intrusive to keep `GetHomeDir()` and drop `GetStateDir()` instead, as I suggested in #493.

freepotion:
@iology [wrote](https://github.com/Attnam/ivan/pull/490):

> GetHomeDir is also a confusing function name on unix systems.

For me, both names are equivalent. I can do it:

> It would have been less intrusive to keep GetHomeDir() and drop GetStateDir() instead, as I suggested in #493.

@iology @fejoa What is your opinion? :alien:

akien-mga:
If `GetHomeDir()` is considered confusing (I tend to agree), then maybe `GetUserDir()` would be better?

iology:
I submitted another pull request: #496

I unify those two functions with `GetUserDataDir()`, to contrast with `GetDataDir()`.

"./" is no problem on Windows XP. (9x not tested; DOS version has been long ignored and not maintained)

freepotion:
Thanks, that's great! :tada: I think I can close this PR now. The discussion can be continued in #496  if necessary.

PR #498, opened 2018-11-21
Update INSTALL with solution to SDL1 and SDL2 clash

fejoa:
Build system bugs if you have SDL1 and SDL2 installed at the same time. See: https://attnam.com/posts/29409 for the type of error and https://attnam.com/posts/29417 for the workaround

Probably needs fixing at the cmake level


iology:
This may help: `export SDL2DIR=/path/to/SDL2/SDL/include:/path/to/SDL2/SDL/lib`

PR #501, opened 2018-11-24
New items

red-kangaroo:
Any comments, thoughts and objections greatly appreciated.


AquariusPower:
Some of the commit's comments should/could be in the .dat files, like the cool one about the "artifact mask" vs mindworms, but I was looking at item.dat and saw no field from which such text could be converted into a in-game item description/history. 
One day (another issue/PR here) may be **the Attnam librarian could tell stories about unique items and such texts could be his words as soon we ask him to identify them for a price?** I mean that instead a non immersive instant description :).
So, such texts could be stored at item.dat in a new field to be used later one day, I mean to not "lose" them in the commits.
Unless I am missing something..

> troll blood give regeneration instead of instant healing, to
differentiate it from healing potions

always wondered why the same thing in both potions, now it looks good :)

> Make boots of kicking always knock back

nice!

> Mana decreases cooldown of magic items

interesting!

Main/Source/miscitem.cpp
```diff
@@ -4015,3 +4019,20 @@ alpha mangoseedling::GetOutlineAlpha(int Frame) const
...
+col16 skeletonkey::GetOutlineColor(int Frame) const
+{
+  switch((Frame&127) >> 5)
+  {
+   case 0: return BLUE;
+   case 1: return GREEN;
+   case 2: return RED;
+   case 3: return YELLOW;
+  }
+}
```
fejoa: `return TRANSPARENT_COLOR;`

Main/Source/miscitem.cpp
```diff
@@ -4015,3 +4019,20 @@ alpha mangoseedling::GetOutlineAlpha(int Frame) const
...
+
+alpha skeletonkey::GetOutlineAlpha(int Frame) const
+{
+  Frame &= 31;
+  return Frame * (31 - Frame) >> 1;
+}
+
+col16 skeletonkey::GetOutlineColor(int Frame) const
+{
+  switch((Frame&127) >> 5)
```
fejoa: if(!IsBroken())

Main/Source/miscitem.cpp
```diff
@@ -4015,3 +4019,20 @@ alpha mangoseedling::GetOutlineAlpha(int Frame) const
...
 }
+
+alpha skeletonkey::GetOutlineAlpha(int Frame) const
+{
+  Frame &= 31;
+  return Frame * (31 - Frame) >> 1;
+}
+
+col16 skeletonkey::GetOutlineColor(int Frame) const
+{
```
fejoa: if(!IsBroken())

Main/Source/miscitem.cpp
```diff
@@ -4015,3 +4019,20 @@ alpha mangoseedling::GetOutlineAlpha(int Frame) const
...
+  switch((Frame&127) >> 5)
+  {
+   case 0: return BLUE;
+   case 1: return GREEN;
+   case 2: return RED;
+   case 3: return YELLOW;
+  }
```
fejoa: `return TRANSPARENT_COLOR;`

Main/Source/miscitem.cpp
fejoa: Although keys don't break, this is a useful way to create a default return value so that clang C++ compilation works.

iology:
> Make boots of kicking always knock back

Oh my ass... How about just a high probability for the knock back effect? The kicking skill can make it 100%.

> Artifact pick-axe that's unbreakable and can be zapped once per hour to create walls.

Should it be cursed?

red-kangaroo:
@AquariusPower I was thinking about adding some item descriptions, ideally for all items. IDK where they should show, though.

@idology You need to connect the kick for the knockback, which already requires some skill. I would be hesitant to make it even less likely, given how few people will have good kicking skill.

Cursed how?

iology:
Idea: The unpick must be wielded when you zap it (auto-wield). The magical power is unholy and has a chance to curse the pick.

red-kangaroo:
I'll note it down, todo once we have blessed/uncursed/cursed subsystem for items. :)

fejoa:
> IDK where they should show, though.

@red-kangaroo maybe the `F1` key once it is implemented?

AquariusPower:
ex.:
```
lantern
{
  NameSingular = "lantern";
  Description = "all the cool text go here"
```

I think just adding the `Description` field at `item.dat` could run w/o crashing (not sure tho).
The implementation could be like was done at `cmdcraft.cpp`:
```
void addRecipe(recipe* prp){
...
  craftRecipes.AddEntry(festring()+prp->action+" "+prp->name, LIGHT_GRAY, 20, prp->iListIndex, true); 
  craftRecipes.SetLastEntryHelp(prp->desc);
```

To simplify, instead of modifying AddEntry() I just let the list allow setting it to the last added entry with SetLastEntryHelp(). To recode to use AddEntry() would take some time and be "complicated" but would be better than this "easy trick" (in the other hand this trick could be more widely used to avoid too complicated methods on lists).

AquariusPower:
@red-kangaroo the capability to use F1 on lists is already merged, doesnt depend on crafting PR :)
https://github.com/Attnam/ivan/blob/master/FeLib/Source/felist.cpp#L1113

fejoa:
Can we please resolve conflicts so we can merge this?

PR #502, opened 2018-11-25
Fix heap-use-after-free in go::Handle()

iology:
The game crashed randomly on mac.


fejoa:
Every bug fix is appreciated by everyone. Thanks :)

PR #506, opened 2018-11-30
Add thievery

red-kangaroo:
As per [this](https://attnam.com/topics/Stealing-from-shops) idea, allow stealing items from shops with a certain chance. [Here](https://docs.google.com/spreadsheets/d/1tKkgHAEDplVN6bnq4emM9OSOvEtNqeIwWYCv9_tLoZY/edit?usp=sharing) is the formula used.


AquariusPower:
I read there, interesting,
may be instead of starting a full fight, we would just be hit until dropping the item inside the room AND leaving the room, so the item(s) would have to be marked as stolen and all of'em dropped would allow ending the fight.
Coming back to the room would make the shopkeeper cry and run to hit the player, but only he, not the guards, just to let we remember.
The only way to buy things again there would be to remain polymorphed!

But of course these things could be another PR later :)

Issue #491, opened 2018-10-18
Cmake option to turn off backtrace?


travankor:
I tried to compile IVAN on musl libc (Linux) and it didn't work since execinfo is glibc-specific (not POSIX). For people that don't need to do debugging, it would be nice if the backtrace could be turned off as a cmake option.

Alternatively, maybe something like libunwind can supplement/replace execinfo.   https://www.nongnu.org/libunwind/

AquariusPower:
The same way we use the -DWIZARD seen at INSTALL info, we could have -DBACKTRACE (I think -DDEBUG may conflict with something more generic but I may be wrong).

Then, instead of ex. `#ifndef WIN32` wherever `backtrace()` is called, we could instead `#ifdef BACKTRACE`.

Do you think you could propose (code/test) such a patch (PR)? 
Basically condition everywhere that is coded `#include <execinfo.h>` to such define too, then catch everywhere it is required :)

It would also be good to be sure about other possible specific requirements by `musl libc`, I mean, not only about `backtrace()` would such PR be.

PR #510, opened 2018-12-06
Backtrace

travankor:
closes https://github.com/Attnam/ivan/issues/491

The current backtrace function is glibc-specific so it will only work on specific GNU/Linux and GNU/Hurd systems. This PR moves the glibc-specific backtrace to a compile-time flag. 

I successfully compiled and played ivan on a musl libc Linux system with these patches.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,7 @@ add_definitions(-DIVAN_VERSION="${PROJECT_VERSION}" -DUSE_SDL)
 
 option(BUILD_MAC_APP "Build standalone application for MacOS" OFF)
 option(USE_HOME_FOR_STATE_DIR "Statedir will be /.ivan/ in current user's homedir" OFF)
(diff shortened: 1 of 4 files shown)


iology:
can also add an cmake option for wizard mode, so: `-DWIZARD=ON` `-DBACKTRACE=ON`

for `-DBACKTRACE=ON`, can first check if "execinfo.h" exists. Then leave the CXX_FLAGS for other workarounds.

travankor:
to be honest, I've never written cmake files before. Any advice how to do this?

iology:
It will take not a small amount of time to learn cmake, so I just pushed a few commits to your branch:

* auto-enable backtrace & no cmake option
* provide a cmake option for wizard mode

Please check if it still compiles well. Thanks~

fejoa:
Is this ok to merge? @iology @AquariusPower ?? :)

iology:
should be no problem.

PR #514, opened 2018-12-14
Fix crashes about recipe::whereRaw()

iology:
e.g. cut web and answer the question by pressing the spacebar


andrewtweber:
Can you check Travis CI? Sometimes the builds will fail but they usually work after a retry. But I've retried that one 3 times and it keeps failing.

fejoa:
I saw the same thing, it's just the mxe compiler toolchain, the package server is down for some reason. I opened an issue here: https://github.com/mxe/mxe/issues/2253

If the other builds are happy then it is usually all good to merge. I think this contribution can be merged, along with #501

Issue #516, opened 2018-12-16
Cutting web crashes if stuck on it


AquariusPower:
```
#0  character::TryToUnStickTraps(this = 0x39f84f0, Dir = ) at Main/Source/char.cpp:11630
#1  character::Move(this = 0x39f84f0, MoveTo = , TeleportMove = false, Run = false) at Main/Source/char.cpp:1126
#2  character::TryMove(this = 0x39f84f0, MoveVector = , Important = true, Run = false, pbWaitNeutralMove = 0x7ffffffc2a90) at Main/Source/char.cpp:1467
#3  character::GetPlayerCommand(this = 0x39f84f0) at Main/Source/char.cpp:3491
#4  character::Be(this = 0x39f84f0) at Main/Source/char.cpp:1087
#5  pool::Be() at Main/Source/pool.cpp:31
#6  game::Run() at Main/Source/game.cpp:1112
#7  main(argc = 1, argv = 0x7ffffffc2e58) at Main/Source/main.cpp:164
```

At character::TryToUnStickTraps()
The field `character::TrapData` is not NULL when it should be after using the craft cut web functionality in case the player is stuck on it, selecting direction as '.'
That linked list first entry `character::TrapData` should be NULL after success on cutting the web but it is not.

**This seems a tricky bug that may require some hours of experimentation/debugLogReading to find out, because why `character::TrapData` is not NULL? is it being set again AFTER web cutting works? or for seme reason web cutting is not cleaning it?**

fejoa:
> the player fumbles and is moved to the web square

Which line of code is that? Is it here: https://github.com/Attnam/ivan/blob/d10e677d3fc68475689f4fe8fe3ea6bfd2941592/Main/Source/cmdcraft.cpp#L1279 ?

AquariusPower:
@fejoa
yes that's it, as I remember I used the teleport code as reference, but now I think I should have used something more robust like what is probably used on the moveTo code when we use the direction keys

@red-kangaroo 
may be FallTo() could be the perfect solution, by it's name feels like a forced moveTo?

Also, it would be good if a check for already stuck on a web is performed before getting stuck on a new web, but... it may be too much precision, and the way it is now "navigating thru webs" could be a fun thing actually lol

red-kangaroo:
@AquariusPower 
> it's name feels like a forced moveTo?

Yeah, that's exactly what it should do. Also there is a IsStuck() function if you want to check whether the player is trapped in a web/beartrap before trying to move them.

Issue #481, opened 2018-08-29
c++ code auto-formatting template/tool needed


AquariusPower:
https://github.com/Attnam/ivan/pull/478#issuecomment-416783442

basically, we dont need to code in a new style we are not used to
as long we have a good auto-formatter template (create one on EclipseIDE or some command line that runs in all OSs to be run just b4 cmake)

iology:
I'm definitely not a good C/C++ programmer. I see people use "[astyle](http://astyle.sourceforge.net/)" or "clang-format". The former require less dependencies and is easier to install.

emlai:
+1 for clang-format. I think it's the way to go nowadays.

PR #382, opened 2018-06-01
(ready) Crafting

AquariusPower:
~~**To test, prefer this branch: https://github.com/AquariusPower/ivan/tree/THE_MERGED_BRANCHES_I_AM_PLAYING , I will avoid duplicating code as much as possible between branches to lower maintenance difficulty, this means, if it is not compiling here is because depends on other PR branches :)** mainly this one: https://github.com/Attnam/ivan/pull/458 , may be I should just merge it here? I can still keep that PR, the merging is easy thru `git gui`~~ that PR is now merged here, so that it compiles properly, and that is a pre-requisite now

First steps on recipe/crafting code. Let's say it is being designed yet.

Any kind of thing is craftable (spawnable), the whole point is balance and challenge.

The recipe system is just a scratch, but usable. This means other recipes can be added.
Good would be to let ingredients requirements count be reused to provide help info on the list, preferably creating a struct to hold that data.
May be an external .dat wouldnt be a good thing, as allowing users to customize them could become too easy, so hardcoded may be better? or it will just hinder creativity?

but... **I am not sure crafting is a good thing anyway**, it is just fun, but if it makes things too easy, will take the fun out of the game. 
TODO: it seems easy to code the action that will spend time in turns to craft things.
May be this system can sit there disabled until we think it is not game breaking.

### Chair Recipe (still unbalanced and easy, just as basic test)
you can now build a chair anywhere, as long you have 20 sticks of anything.

is that recipe balanced? 

whenever you get a chair you will be able to heal faster, 
w/o this recipe you are required to travel to other dungeon levels or even outside it, 
so may be we should remove this recipe later or increase it's constraints? 
like where it can be placed, like only if near a furnace or may be a fountain?
add required ingredients, may be spider web or nails?
make it be more fragile than other chairs?

Problem0: I didnt limit it to fir sticks as configured in the .dat file.

Problem1: it currently is instantaneous, an action class instance (like dig) is required to make it take time and be interruptible (and preferably continuable), and it's data will be put in the savegame file also.

Possible Improvement: wooden walls could drop more rare planks and these be required instead of sticks?

### Wall recipe
made of stones

### poison
from spiders 100ml per corpse

### other stuff
many other things were added, just try the game and hit F1 for each recipe help :)

PR #517, opened 2018-12-16
Cutting a web will now restore the previously wielded item.

AquariusPower:
will still only spend one turn as this action is like a single slash on the web

and now you will not see/know what tool/weapon was used (TODO unless we message log it "You use %s to try to cut the nasty web."), there may have some complicated way to see the wielded being changed tho :>


iology:
btw should have a notice for lost weapons, otherwise it looks like the bug is not fixed.

AquariusPower:
> btw should have a notice for lost weapons, otherwise it looks like the bug is not fixed.

But... it is already there
```
if(bLoseWeapon){
        rpd.itTool->RemoveFromSlot();
        rpd.itTool->MoveTo(rpd.lsqrPlaceAt->GetStack()); //TODO check if is not a WALL!!!
        ADD_MESSAGE("I lost my %s!",rpd.itTool->GetName(UNARTICLED).CStr());
        rpd.bAlreadyExplained=true;
}
```

Also the most important bug is actually the annoyance of having to swap weapons back to what it was, that breaks the game flow and makes players go away.

The important thing when coding, if you have not all the time of the world, is: **priorities**.
If we have no priorities we will spend a lot of time doing **minor** things other than crucial ones, we should spend time on minor things only **after the crucial** ones are fixed/working and **after we have dealt with our own life matters**.
So basically, to show the weapon when cutting a web makes no difference to the gameplay mechanics, therefore it is a **minor** thing. It can be considered a bug but certainly **not a crucial one**.

That's why I urge for ppl willing to help on coding. 
More users, that are developers too, interested on the project was always my main goal, that's why I created the "console commands" and provided the debug messages project, but it being in BSD3 I didnt commit it here, but I think it can be committed like was done with some other libs right? 
Such step, if someone could prepare, would help even more new devs to come by! So more hands to take care of tasks other devs have no time to do! ;)

AquariusPower:
> Just tested. It seems this problem only happened when I wielded two weapons and one of them is a dagger (on left or right hand).

as to it happen is random (and somewhat rare), it would be better to temporarily change the code to let it happen always, to be sure about the origin of the problem spending little time on it's simulation.

> The new BSD-3-clause license should be compatible with GPL

so anyone can put dbgmsg files into a new folder like was done with xbrzlib,
prepare a CMakeLists file and it will be ready for new devs to use it,
I still wonder if there is a better/easierToUse debug log lib out there with more features and performance,
otherwise dbgmsg would still suffice for now.

I think I am going to need a good memory brain-implant one day, better save for it, thx xD

> It should be a good habit to remove unimportant trailing spaces, so when people fix others bugs they won't accidentally modify another unrelated files/lines, thus reducing merge conflicts. Or we should tell everyone not to strip trailing spaces, but if someone doesn't care, then again it's likely to produce merge conflicts. We'd better not leave the baggage to the future contribution. I know this is not a serious issue right at the moment but there is the potential.

interesting, if I am not wrong, most of my last conflict solving were about  things that could have been automatic out of it's lack of complexity, but I may be asking too much or may be github's merging tool configurations could be tweaked by us end users (I dont know if we even have access to such tho)?

I particularly think we should not annoy ppl willing to contribute with their **spare time, ideas and energy, for free**, to projects they love, because of **minor things**, or what will probably happen is that **ppl will just create and distribute their forks** (could that be what happened here in the past? if so cant we prevent it from happening again?) to escape from being bossed around like they are in their most probably oppressive and low paying jobs everywhere in this non utopic world, like in that they could be at parties having fun and many endless pleasures but they are sat behind their bright monitors, with a noisy fan nearby, focusedly coding to help us improve such marvels! 
Aww dammit! Did I just woke up some potential contributor :o ?

But do not despair my friend! It will never be a serious issue if someone could spend some time on preparing a good [C++ coding style template formatter](https://github.com/Attnam/ivan/issues/481) to be run before every compilation, it would be amazing! :)

iology:
Man, please calm down. Other than you, nobody can set a deadline for your great work here.

If stripping trailing whitespace is not simple enough for you, fair, someone may do that for you in other pull requests.

Also, anyone can merge your code. This is only a conflict between you and me. My personal standard is just a little bit higher, at least before a true leader (boss?) appear.

My first priority is to play IVAN and to eliminate crashes. I'm sorry for reporting too many issues and not putting my hands directly on improving your code. Everyone's time is valuale.

Others please take care of this pr.

AquariusPower:
And I would never be able to properly fix `festring` neither make it work on Mac! much thx! :)

Don't understand me wrong also, I think it is great to have someone that likes to have all the code as perfect as possible, but I have a lot of difficulty doing that, one of the reasons I left my old job lol.

Also, I dont think the reports to improve this game should stop. Anyone who can spot things to be improved should do so! And anyone that have spare time could give us a hand also xD.

PS.: I was thinking, it seems we have quite a lot of open issues, I could even guess some already deprecated after so many changes, for ex. at the `entt` project the issues are closed after a few days apparently just to keep it clean, we could close all issues older than 6 months just because they are old, and if needed who created them just re-open with a new message and it will be refreshed. Just my opinion any way :)

fejoa:
I'm not very good in these "grr" situations, I hope there's no hard feelings between you two.

I'm going to merge this PR anyhow, since it improves the user experience. The crash bug still needs fixing but that has been noted in #516 

@AquariusPower I like your contributions so far, so keep it up. Upon reading your code I have found two ways your C++ coding could be improved:

1. Avoid the use of #defines where a simple function would work fine. Preprocessor directives are ok, for example in [whandler.cpp](https://github.com/Attnam/ivan/blob/master/FeLib/Source/whandler.cpp#L706), but they came up in [cmdcraft.cpp](https://github.com/Attnam/ivan/pull/382/files#diff-7fbccbc9b4f87c5e177faf8526c3bd4eR1217) and they just made the code a bit harder for humans to read.

2. The naming of your variables could be more descriptive. The legacy code we have does this to good effect. When I read lines like [this one](https://github.com/Attnam/ivan/pull/382/files#diff-7fbccbc9b4f87c5e177faf8526c3bd4eR1295), I found myself needing to scroll up to figure out what the variables were referring to. 

On the whole, your C++ is continually improving and I especially appreciate your emphasis on stable code, as evidenced by the debug tools you have written, and the way you alert others to bugs. It is true that the original devs coded many bugs, which we occasionally still find even today, and that did not stop them from releasing their work.

@iology you have rightly identified a significant gap in this project's organization, namely that we have no [contribution guidelines](https://github.com/Attnam/ivan/community). A coding style guide has been [called for in the past](https://attnam.com/posts/25604), and this could be a good starting point.

iology:
What I am angry about is: apparently, the time&energy spent on crashes, testing and code review is overlooked & who are willing to be responsible for hastily merging the OBVIOUSLY buggy code? Not me, and there is no deadline for either side so everyone has time to relax, make plans and do a better job instead of not trying to be a better and responsible developer.

I hope this pr will not caused any serious bug when I do not use those features. The coding style cannot solve the root problem.

Last, I just dislike the attitude, not anyone. I'm sure we can have more ways to cooperate more harmoniously if we are coding together in a cozy room, but, well... time to relax & congrats on Ivan055

PR #527, opened 2019-01-19
Improve path search for SDL2

iology:
This should resolve #498 if both SDL1 and SDL2 exist and their include paths are `/prefix/include` (not `/prefix/include/SDL`) and `/prefix/include/SDL2`.


fejoa:
Excellent!

PR #528, opened 2019-01-20
Doc: build IVAN in a custom folder

iology:
The travis ci has been doing this for a while.

btw it is time to discard SDL 1.x.


AquariusPower:
> btw it is time to discard SDL 1.x.
and probably DJGPP too?

iology:
+1 I don't understand why people are so silent when talking about deprecating the DOS build, since the old code is always in the git history for the enthusiasts.

PR #532, opened 2019-02-15
Further restrict the range of craftable items

iology:
esp. prevent making a magical item made of plant fiber

### DIFF
--- a/Main/Source/cmdcraft.cpp
+++ b/Main/Source/cmdcraft.cpp
@@ -129,14 +129,20 @@ bool craftcore::canBeCrafted(item* it){
   if(dynamic_cast<lump*>(it)!=NULL)
     return true;
 
-  const itemdatabase* itdb = it->GetProtoType()->GetConfigData()[0];
+  const itemdatabase* itdb = it->GetDataBase();
 
+  // TODO figure out how to create an empty can or bottle
   if(
     game::IsQuestItem(it) ||
     it->GetEnchantment()!=0 ||
     it->GetCategory()==BOOK ||
     it->GetCategory()==FOOD || //TODO allow near oven, but the problem is the ingredients... also may be blocked below too
+    it->GetCategory()==MISC ||
+    it->GetCategory()==SCROLL ||
+    it->GetCategory()==POTION ||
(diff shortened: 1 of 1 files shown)


AquariusPower:
and also made of anything else,
wish I knew that b4 creating `cmdcraftfilters.cpp` and related script `prepareCraftFilters.sh`, wonder if they are still useful? :>

to see almost a full list of what will be further denied (this is not the best filter tho)
`egrep "PostFix = \"[^\"]|Possibility = 0| Config |^[^ {}]" item.dat`

iology:
You can get a lot of "plant fiber stones" in UT1, then make a lot of belt of levitation and fly across the ocean.

I also blacklisted the categories: POTION, SCROLL, MISC.

A better approach could be to add the property "canBeCrafted" in Script/item.dat . The property can be inherited from the base class. This can be done after the planned release in this month (but I'm not sure if that will spoil my fun ;).

iology:
Or a property "CraftDifficulty": 0 - impossible to craft, 1 - easiest, 2,3,...,N - harder

red-kangaroo:
I like CraftDiffifulty. All magic items should be uncraftable anyway.

AquariusPower:
at `cmdcraft.cpp` there is `applyDifficulty()` with some stuff already configured (hard coded) there, I think such database field could fit directly there at `recipedata.fDifficulty`.

**The fumble formula:**
it is a per turn check for fumble, imagine like each hammer/carvingTool strike, may break the thing,
crafthandle::CheckFumble() calls craftcore::CheckFumble()
Crafting should not be something very difficult.
As initial stats are all around 10, to require some progress to become a "novice crafter", it would require to have all stats around 20 (there is another formula for each stat weight) but I am not sure that is a good value, could be more than 20 I guess?
Btw.: The minor fumble result (weak explosion) was just a fun/quick/easy way to let something bad happen (fire sparks from meltables at anvil), but makes no sense when carving on a workbench.

fejoa:
This seems like a self-contained change. Can it be merged now?

iology:
Yes!

PR #536, opened 2019-02-21
Give more installation instruction about the mac app

iology:
Not very confident in my English so please do some review. ;)


andrewtweber:
Just a tiny improvement to the first sentence:

> To install, simply drag IVAN.app into the Applications folder. Then
it will also appear in Launchpad.

iology:
@AquariusPower I rolled back to 93efe67 and removed the restriction, but could not reproduce the problem. So it probably had been fixed sometime.

iology:
Thanks~

PR #537, opened 2019-02-22
Fix ring detection to work same as scrolls

andrewtweber:
Scroll of detect material was fixed to be able to detect carrot, etc. but the ring of detection was still forbidding those materials.


iology:
The code for ring and scroll are almost the same so I don't see any problem.

PR #538, opened 2019-02-24
Wish for empty containers & Make empty cans and empty bottles craftable

iology:
Also add aliases for land mines and vials. Before this change it was impossible to wish for a bottle using text "potion" or "drink".

### DIFF
--- a/Main/Source/cmdcraft.cpp
+++ b/Main/Source/cmdcraft.cpp
@@ -131,20 +131,36 @@ bool craftcore::canBeCrafted(item* it){
 
   const itemdatabase* itdb = it->GetDataBase();
 
-  // TODO figure out how to create an empty can or bottle
   if(
     game::IsQuestItem(it) ||
     it->GetEnchantment()!=0 ||
     it->GetCategory()==BOOK ||
-    it->GetCategory()==FOOD || //TODO allow near oven, but the problem is the ingredients... also may be blocked below too
     it->GetCategory()==MISC ||
     it->GetCategory()==SCROLL ||
-    it->GetCategory()==POTION ||
     !itdb->CanBeWished ||
     itdb->Possibility <= 0 ||
     !itdb->PostFix.IsEmpty() ||
     false // just to make it easier to re-organize and add checks above
   ){
+    return false;
+  }
+
+  if(it->GetCategory()==POTION)
+  {
+    item* Bottle = dynamic_cast<potion*>(it);
+    if(Bottle && !Bottle->GetSecondaryMaterial() && Bottle->GetNameSingular() == "bottle")
+      return true;
+
+    return false;
+  }
+
+  // TODO allow near oven, but the problem is the ingredients... also may be blocked below too
+  if(it->GetCategory()==FOOD)
+  {
+    item* Can = dynamic_cast<can*>(it);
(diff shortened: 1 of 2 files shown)


andrewtweber:
I believe that before, if you specifically wished for a "bottle of ommel urine" for example, it would just give you a randomly-filled bottle, is that behavior changed?

fejoa:
Looks great! I think it just enables wishing of empty bottles etc. by the addition of the keyword 'empty', in addition to allowing empty cans and bottles to be crafted.

iology:
Yes, still no way to wish for an item with arbitrary material, unless we allow keywords like "of ommel urine".

I found that `Config FEATURE` or some thing still invalidates some aliaes for matching (not my fault). You can search a broken lantern by "broken lantern" but not by "broken light". Not so practical but I haven't figured out why...

iology:
Ah, found out the cause: itemdatabase::InitDefaults() automatically adds the adjective "broken" to every BROKEN item.

For convenience, I think the container should be empty by default for crafting cans and bottles.

andrewtweber:
1 conflict appeared after merging the other PR. Feel free to merge after you resolve it @iology

AquariusPower:
As I remember, 
before cans where blocked by being filled with food (therefore preventing crafting'em empty too in the old code), 
if we craft them using any other material than the default one (tin?), 
we can fill'em with liquids,
but after using it's contents, it would **spawn an empty can with the default material** and not the original one,
I guess it may still be like that,
it was one of the reasons not much time was spent on letting it work at that time, out of that inconsistency :>

iology:
Just fix some logic there, does the problem still exist?

* Wish for a can of liquid -> change material -> drink
* Craft empty can -> fill liquid -> drink

iology:
@AquariusPower I rolled back to 93efe67 and removed the restriction, but could not reproduce the problem. So it probably had been fixed sometime.

AquariusPower:
I just tested and the problem **is not happening**: I changed can's material to gold, consumed it's contents and the empty one was still gold.

To be sure, prayed to mellis and the golden emtpy can become a filled one but now made of default material, the god doesnt fills the existing can, but really trades it by a new one, as it is said there. I dont know if it could instead fill'em (to be more fun?), but is at least coherent with what is said :>

iology:
> To be sure, prayed to mellis and the golden emtpy can become a filled one but now made of default material, the god doesnt fills the existing can, but really trades it by a new one, as it is said there. I dont know if it could instead fill'em (to be more fun?), but is at least coherent with what is said :>

`item::BetterVersion()` is only used by Mellis. Maybe we should move it to gods.cpp. +1 for a new one.

AquariusPower:
mmm.. nice spotting

at Main/Source/miscitem.cpp

```
item* can::BetterVersion() const
{
  if(!GetSecondaryMaterial())
    return can::Spawn();
  else
    return 0;
}
```

could instead use self main material config at this call, I think like this: `return can::Spawn(GetConfig());`  
whenever explosions and ennerbeast be able to xplode the can (cant right?), crafting better quality ones will be interesting

potions/bottles seem to already work like that (dont understand why tho, gotta test it to confirm)
`return potion::Spawn(GetConfig());`

iology:
`GetConfig()` only consider the material config in item.dat, so special materials won't be kept.

Issue #535, opened 2019-02-20
Animation does not work properly


red-kangaroo:
Two players reported animation (smoke and flies) and menu transition not working on Windows release.

> win8.1; like i said, frameskip fixes the problem with resting, but things that have animations outside of the regular turn are bugged i think now, like fires burning, flies on rotting organic materials, lights, etc. it only advances frames when you press an input, the menu has a similar issue as well (fade out of the option menu only advances as you press a key, isnt automatic)

> Same issue with resting not advancing game state without input from player on my end, Win10 and no frame skip. Screen hangs midway into the transistion to main menu, but input still works fine

Cannot reproduce on Linux, my build seems to work properly.

fejoa:
Thanks for bringing this up, I've noticed it too. Curses. I wonder what might be causing it?

AquariusPower:
even on linux, frameSkip=-2 at least seems a bit inconsistent, it will only "not animate" (wait) some times, other times it behaves apparently like -1.

by what he describes, he **could** be using -2 (wait),
but what he expects is -1 (auto) and >=1 (custom fixed frameSkip),
I guess he is using -1
I wonder if he use >=1, may be 3, could still let it work well until some coder using windows OS could track it?

iology:
Exactly what are your config values for FrameSkip?

AquariusPower:
I think everyone is using the default automatic -1.
I would like to wild guess that, if you increase the ivan's process priority on windows OS to real-time if it would make any difference? could be something outside the code itself too.
May be, some coder on linux/mac could try to run the windows executable thru Wine or VMWare too?

PR #546, opened 2019-03-04
Don't drop containers after tasting unless empty

### LINKED ISSUE #545: Tasting liquids will drop the partially full bottle
If you have the "Drop food leftovers automatically" config option enabled, it will drop the bottle after you '**T**'aste it, even if the bottle is mostly full.

andrewtweber:
Might fix #545. My SDL is broken so I can't compile and test yet.

### DIFF
--- a/Main/Source/actions.cpp
+++ b/Main/Source/actions.cpp
@@ -161,7 +161,15 @@ void consume::Terminate(truth Finished)
 
   if(Finished)
   {
-    if(Consuming->Exists() && !game::IsInWilderness() && (!Actor->IsPlayer() || ivanconfig::GetAutoDropLeftOvers()))
+    truth PlayerWantsToDiscard = false;
+    if(Consuming->GetSecondaryMaterial()){
+      PlayerWantsToDiscard = (
+        ivanconfig::GetAutoDropLeftOvers() &&
+        (Consuming->GetSecondaryMaterial()->GetVolume() == 0) // don't drop after tasting
+      );
+    }
+
+    if(Consuming->Exists() && !game::IsInWilderness() && (!Actor->IsPlayer() || PlayerWantsToDiscard))
     {
       Consuming->RemoveFromSlot();
       Actor->GetStackUnder()->AddItem(Consuming);
--- a/Main/Source/actions.cpp
+++ b/Main/Source/actions.cpp
@@ -161,19 +161,23 @@ void consume::Terminate(truth Finished)
 
   if(Finished)
   {
-    truth PlayerWantsToDiscard = ivanconfig::GetAutoDropLeftOvers();
-    if(Consuming->GetSecondaryMaterial()){
-      PlayerWantsToDiscard = (
-        ivanconfig::GetAutoDropLeftOvers() &&
-        (Consuming->GetSecondaryMaterial()->GetVolume() == 0) // don't drop after tasting
-      );
-    }
-
-    if(Consuming->Exists() && !game::IsInWilderness() && (!Actor->IsPlayer() || PlayerWantsToDiscard))
+    if(Consuming && Consuming->Exists())
     {
-      Consuming->RemoveFromSlot();
-      Actor->GetStackUnder()->AddItem(Consuming);
-      Actor->DexterityAction(2);
+      truth PlayerWantsToDiscard = ivanconfig::GetAutoDropLeftOvers();
+      if(Consuming->GetSecondaryMaterial())
+      {
+        PlayerWantsToDiscard = (
+          ivanconfig::GetAutoDropLeftOvers() &&
+          (Consuming->GetSecondaryMaterial()->GetVolume() == 0) // don't drop after tasting
+        );
+      }
+
+      if(!game::IsInWilderness() && (!Actor->IsPlayer() || PlayerWantsToDiscard))
+      {
+        Consuming->RemoveFromSlot();
+        Actor->GetStackUnder()->AddItem(Consuming);
+        Actor->DexterityAction(2);
+      }
     }
   }
   else if(Consuming && Consuming->Exists())
--- a/Main/Source/actions.cpp
+++ b/Main/Source/actions.cpp
@@ -161,7 +161,7 @@ void consume::Terminate(truth Finished)
 
   if(Finished)
   {
-    truth PlayerWantsToDiscard = false;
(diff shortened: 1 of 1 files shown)


Main/Source/actions.cpp
```diff
@@ -161,7 +161,15 @@ void consume::Terminate(truth Finished)
...
-    if(Consuming->Exists() && !game::IsInWilderness() && (!Actor->IsPlayer() || ivanconfig::GetAutoDropLeftOvers()))
+    truth PlayerWantsToDiscard = ivanconfig::GetAutoDropLeftOvers();
+    if(Consuming->GetSecondaryMaterial()){
+      PlayerWantsToDiscard = (
+        ivanconfig::GetAutoDropLeftOvers() &&
+        (Consuming->GetSecondaryMaterial()->GetVolume() == 0) // don't drop after tasting
+      );
+    }
+
+    if(Consuming->Exists() && !game::IsInWilderness() && (!Actor->IsPlayer() || PlayerWantsToDiscard))
```
iology: `Consuming->Exists()` seems unnecessary because if `game::SearchItem()` succeeds the item's memory is not yet freed. Should check `Consuming != NULL` instead.

`entity::Exists()` is a really confusing function.

iology: Need to check both: `Consuming && Consuming->Exists()`, before calling other member functions.

andrewtweber: @iology good call, I wrapped everything inside a `Consuming && Consuming->Exsists()`

andrewtweber:
I think there may be a problem with compiling on OSX. I've tried on both my work and home computers. The game compiles but then the main menu screen is just plain black. Didn't used to have this problem.

Can somebody please compile and test my PR? And "squash" when you merge.

Test cases that I wanted to try but can't:
- Drinking full bottle & can (should drop empty)
**- Tasting full bottle & can (should not drop)** - this is the one that I'm trying to fix
- Tasting last bit of bottle & can (should drop empty)
- Eating food with leftover (e.g. banana, should drop peel)
- Eating food without leftover (e.g. kiwi)

fejoa:
> Can somebody please compile and test my PR? And "squash" when you merge.

I compiled your branch and ran it. I get the normal menu screen on boot-up so I'm unable to reproduce the black screen error. I do however observe more of the problems described in #535. Like if I save the game and flee, I get a glitchy transition back to the menu. I wonder if it is related...

andrewtweber:
@fejoa were you able to test the taste fix?

iology:
All tested.

Is the pre-compiled sdl libraries from Homebrew not compatible with your macos?

andrewtweber:
I have no idea what's going on with SDL or ivan. When I run on my work computer I get the black main menu and this console output:

```
bjc[79620]: Class SDLApplication is implemented in both /usr/local/opt/sdl2/lib/libSDL2-2.0.0.dylib (0x10c7e31c0) and /Library/Frameworks/SDL2.framework/Versions/A/SDL2 (0x10c6b1680). One of the two will be used. Which one is undefined.
```

plus several more similar errors for different SDL classes. When I deleted the files in `/usr/local/opt/sdl2` the game won't run because `Library not loaded: /usr/local/opt/sdl2/lib/libSDL2-2.0.0.dylib`.

But when I delete `/Library/Frameworks/SDL2.framework` the game won't run because `Library not loaded: @rpath/SDL2.framework/Versions/A/SDL2`

So I guess it's using both.... at the same time.....

On my home computer I don't get the above errors but the main menu is still black.

iology:
Looks like an issue about cmake. If a clean build does not help and `otool -L $BIN_PATH_OF_IVAN` shows both libraries, then we can discuss it in a new issue.

andrewtweber:
Yep I have tried cloning the repo fresh, still have the same issue.

Here is the output of `otool`. Kind of weird that SDL is the only one with `@rpath`, not sure what that means?

```
/usr/local/bin/ivan:
	/usr/local/opt/pcre/lib/libpcre.1.dylib (compatibility version 4.0.0, current version 4.10.0)
	/usr/local/opt/sdl2_mixer/lib/libSDL2_mixer-2.0.0.dylib (compatibility version 3.0.0, current version 3.0.0)
	/System/Library/Frameworks/Cocoa.framework/Versions/A/Cocoa (compatibility version 1.0.0, current version 23.0.0)
	@rpath/SDL2.framework/Versions/A/SDL2 (compatibility version 1.0.0, current version 8.0.0)
	/usr/local/opt/libpng/lib/libpng16.16.dylib (compatibility version 52.0.0, current version 52.0.0)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	/System/Library/Frameworks/CoreMIDI.framework/Versions/A/CoreMIDI (compatibility version 1.0.0, current version 69.0.0)
	/System/Library/Frameworks/CoreAudio.framework/Versions/A/CoreAudio (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1560.12.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 400.9.4)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.200.5)
```

iology:
It seems you have sdl2 and sdl2_mixer in homebrew cellar, but not both SDL2.framework and SDL2_mixer.framework in /Library. cmake found SDL2.framework but didn't found SDL2_mixer.framework, so it used libSDL2_mixer.dylib from homebrew cellar which is not compatible with the framework. Solution is to installed SDL2_mixer as well. I'll also add a small check in FindSDL2.cmake.

https://www.libsdl.org/projects/SDL_mixer/

https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/RPATH-handling

https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/RunpathDependentLibraries.html#//apple_ref/doc/uid/TP40008306-SW1

andrewtweber:
On my home computer i was actually missing both frameworks. I installed it and I no longer have the black screen of death! I'll have to check my work computer on Monday

AquariusPower:
> missing both frameworks
and still ran?
sounds like some pre-compilation check could be added 
if I am not wrong (xD) it sdl mixer was added to let the sfx work

andrewtweber:
Thanks @iology , with the  frameworks it's now working on both computers!

PR #547, opened 2019-03-05
Make XT door 5 tiles wide, guards inline

### LINKED ISSUE #541: TX entrance blocked off
Well this is awkward :D Luckily I had a pickaxe and could dig through the house. This could've been game-ending though if I didn't.

An easy solution would probably be to move those guards up one spot to be next to the wall instead of in front of it.

tx

fejoa:
Fixes #541


fejoa:
Closed in favour of #548

PR #551, opened 2019-03-12
Improve item search algorithm

iology:
* Allow more than one space between keywords.
* Fix aliases and reduce search conflicts among items.
* More aliases for searching items.

### DIFF
--- a/FeLib/Include/festring.h
+++ b/FeLib/Include/festring.h
@@ -80,6 +80,7 @@ class festring
   void Empty();
   void Assign(sizetype, char);
   void Resize(sizetype, char = ' ');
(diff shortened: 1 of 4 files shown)


iology:
Because the base item is already named "girdle" and "belt", I checked the commit (39828f3b3f39e50ae9efe8e4a19455487e1ea9ba). Did you mean the thief's girdle should always be named "girdle" instead of "belt"? @red_kangaroo

red-kangaroo:
@iology Yes, the intention was that thief's girdle will always be girdle, while other belts will be either belt or girdle based on material.

fejoa:
Tested on windows, works well. Nice job @iology

PR #553, opened 2019-03-16
Fix #415

### LINKED ISSUE #415: Can't identify rotated piled items
I've been play testing 0.53 a lot, and while I **love** most of the changes, one that I really dislike is the rotated piled items.

Often looking at the pile, it's impossible to tell even what type of item it is. Even more difficult when the items are broken. Example, here is a pair of broken gloves:

phmgni8iesq8lboldlmcqz06f8iwnj_1529270227

I think it was a nice idea in theory but should be reverted

iology:
No rotated pictures for equipments and newly-spawned items.

### DIFF
--- a/Main/Include/item.h
+++ b/Main/Include/item.h
@@ -621,6 +621,7 @@ class item : public object
   virtual void Ignite();
   virtual void Extinguish(truth);
   void SetValidRecipeIngredient(truth b){ValidRecipeIngredient=b;}
+  void ResetFlyingThrownStep() { iRotateFlyingThrownStep = 0; }
  protected:
   virtual cchar* GetBreakVerb() const;
   virtual long GetMaterialPrice() const;
--- a/Main/Source/char.cpp
+++ b/Main/Source/char.cpp
@@ -10964,7 +10964,12 @@ truth character::TakePetItems()
       break;
 
     for(uint c = 0; c < ToTake.size(); ++c)
+    {
+      if(ivanconfig::GetRotateTimesPerSquare() > 0)
+        ToTake[c]->ResetFlyingThrownStep();
+
       ToTake[c]->MoveTo(PLAYER->GetStack());
+    }
 
     ADD_MESSAGE("You take %s.", ToTake[0]->GetName(DEFINITE, ToTake.size()).CStr());
     Success = true;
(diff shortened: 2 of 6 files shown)


red-kangaroo:
Nice!

andrewtweber:
Awesome. So they only rotate if thrown?

iology:
Yes, only when thrown/blown/kicked up. And reset when picked up.

The problem is basically caused by uninitialized values.

AquariusPower:
I would like to add a minor thing you may have perceived (at least prior to this pr):
after some items are thrown, after they rotate, the final rotation is not kept on next turn.
So whenever you step again, that item will "become alive" and 1/4 spin (one spin step) again.

I think the rotation could be reset after thrown, no matter what spin step it was, just to prevent that glitch.
I havent tested this merge yet to see if affected that.

iology:
That isn't fixed yet.

```cpp
  if(ivanconfig::GetRotateTimesPerSquare()>0 && iRotateFlyingThrownStep!=0){ //must be disabled before exiting Fly()
    iRotateFlyingThrownStep=4; //default rotation is w/o the rotation flags at the switch(){}
    //force redraw at default rotation to avoid another spin when player moves TODO how to let it stay in the last rotation?
    RemoveFromSlot();LandingSquare->GetStack()->AddItem(this, false); //TODO find a better way then remove and re-add to same square...
    game::DrawEverything();
    iRotateFlyingThrownStep=0; //disables rotation
  }
```

You mean this? So your solution was already there but the bug made me think the rotation should not be reset.

Now the solution is either to clean up some code from my commit and add this:

```
    while(clock() - StartTime < fFlyDelay * CLOCKS_PER_SEC);
    game::DrawEverything();
```

or remove that line:

```
    iRotateFlyingThrownStep=0; //disables rotation
```

AquariusPower:
```
 while(clock() - StartTime < fFlyDelay * CLOCKS_PER_SEC);
 game::DrawEverything();
```

mmm... I cant recall it, but apparently it worked then, but not as often as it should be (always).
so with that code, it would wait a tiny bit and apply the fix.
if it works, would be great! :)

obs.: I think not disabling the rotation caused other problems.

PR #558, opened 2019-06-02
Nerf magic helmets

red-kangaroo:
Magic helmets with pre-defined materials are too strong, as they appear 
early with good material AND a magical effect.

Make magic helmets use randomized materials. Their magic effects should 
be good enough to make them competitive with other helmets.

Also switch most magic helmets to normal helmet, rather than full 
helmet. This allows full helmets to be more distinct, as they will offer 
the best coverage and AV.

### DIFF
--- a/Main/Source/char.cpp
+++ b/Main/Source/char.cpp
@@ -484,8 +484,9 @@ cchar* character::GetNormalDeathMessage() const
     "done in @k", "defeated @k", "struck down @k", "offed @k", "mowed down @k",
     "taken down @k", "sent to the grave @k", "destroyed @k", "executed @k",
     "slaughtered @k", "annihilated @k", "finished @k", "neutralized @k",
-    "obliterated @k", "snuffed @k", "done away with @k", "put to death @k" };
-  return killed_by[RAND() % 27];
+    "obliterated @k", "snuffed @k", "done away with @k", "put to death @k",
+    "released of this mortal coil @k", "taken apart @k", "unmade @k" };
+  return killed_by[RAND() % 30];
 }
 
 festring character::GetGhostDescription() const
--- a/Script/item.dat
+++ b/Script/item.dat
@@ -5135,7 +5135,6 @@ helmet
                               25, 10, 5, 2, 1, 25, 5; }
     HelmetBitmapPos = 112, 496;
     EnchantmentPlusChance = 30;
-    SoundResistance = 0;
   }
 
   Config FULL_HELMET;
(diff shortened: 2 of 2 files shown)


andrewtweber:
I suppose that matches the other magical items (gloves of dexterity, etc.)

I don't know if I agree with changing them from full helmet though.

fejoa:
Good initiative @red-kangaroo. I agree with @andrewtweber, and will miss magical full helmets. As a compromise, what about splitting them up so that some magical helmets are normal helmets and some are full helmets. Say in the ratio 3:1? That way, the majority of magical helmets will be ordinary ones, while a few will still be full helmets.

@red-kangaroo I think you might have accidentally changed some of the unix file permission modes in https://github.com/Attnam/ivan/pull/558/commits/110565f5824a622ffdc86293c7db3b52a6fab96b
The change stops the Travis virtual machines (really a container running linux) from executing shell script build files on a linux command line (hence the build fails). Can you please try undo the changes by this particular commit and then push to your branch again? You can do the file renaming job in a subsequent commit, or a separate branch if you like.

andrewtweber:
I would be ok with them randomly generating as either normal helmets or full helmets. 3:1 sounds fine

red-kangaroo:
Sorry about the late reply.

I like the idea of splitting the magic helmets among normal and full, but I think @fejoa meant deciding what kinds of helmets would be normal and what would be full, while if I understand @andrewtweber correctly, he means generating all helmets as either normal or full. The latter approach would be rather difficult, IMHO, as the bitmap position, volume, coverage etc. are stored in script file and not saved in the helmet object itself, AFAIK, and thus we would need to rework the code to either decide between two script entries and/or hardcode two values and let them be saved for each helmet object. Alternately, we could declare two different types of each magic helmet (in normal and full appearance), but that would increase the current chances of magic helmet generation, as they are already at 1.

I would instead propose to go with the former (feoja's) suggestion and change some helmets to normal, while keeping others as full. Specifically the more useful stats could be moved to normal helmets while the less often picked would retain their higher AV as full helmets. A bit of balance. :)

For example:
FULL - charisma, will, mana, teleportation
NORMAL - intelligence, perception, telepathy, wisdom

What do you think?

We could also make some magic helmets be magic masks instead, if anyone likes the idea. :)

Will fix the permissions. :)

red-kangaroo:
Sorry, my Git Fu failed me. Let me figure out how to reopen this without the changed files...

fejoa:
Just tickling the virtual machines now...

red-kangaroo:
Travis doesn't like me. :(

fejoa:
It looks like it's just a problem with Homebrew. All the other build configurations passed so your code definitely compiles, and the problem with the file permissions has gone away.

Anyone have any idea why Homebrew needs to install sooo many dependencies?

red-kangaroo:
Changed so that now the stronger magic helmets are normal helmets, while the weaker are full helmets.

fejoa:
> I like the idea of splitting the magic helmets among normal and full, but I think @fejoa meant deciding what kinds of helmets would be normal and what would be full, while if I understand @andrewtweber correctly, he means generating all helmets as either normal or full.

I kinda meant the same as Andrew. However, I agree that to implement that would be needlessly over-complicating things, so let's run with your proposed change as they stands.


> allows the magic helmets to use even cloth materials ( magic hats!) without it looking weird with the full helmets graphic.

What about introducing (for helmets at least) a new bitmap position thingee called `FlexibleBitmapPosition` that works like `FlexibleNameSingular` does for `NameSingular`, in that it would change the bitmap position to a suitable looking "magic hat" graphic if the material is flexible enough??
There would also need to be a `FlexibleBitmapPosition` in the `BROKEN|HELMET` block.

red-kangaroo:
> What about introducing (for helmets at least) a new bitmap position thingee called FlexibleBitmapPosition that works like FlexibleNameSingular does for NameSingular, in that it would change the bitmap position to a suitable looking "magic hat" graphic if the material is flexible enough??
There would also need to be a FlexibleBitmapPosition in the BROKEN|HELMET block.

That might be quite cool.

fejoa:
Ready to merge? I figure we give this change a go.

red-kangaroo:
@fejoa Ready to merge, and I will make a note to see if I cannot change magic helmets into an enchantment that would be possible to apply to either helmet, full helmet or a mask.

PR #561, opened 2019-06-20
Further Golem Balancing

red-kangaroo:
Golem spawning rates are still not OK, with powerful golems sometimes 
spawning way too early and often. This commit tries to address the 
issue. Some playtesting I was able to do seems to suggest that golems
are a bit less likely to spawn too early and made of too strong materials,
but IDK if the issue of golems is truly solved.

Also handle theoretical gaseous golems.

### DIFF
--- a/Main/Include/materia.h
+++ b/Main/Include/materia.h
@@ -164,6 +164,7 @@ class material
   static material* MakeMaterial(int, long = 0);
   virtual truth IsFlesh() const { return false; }
   virtual truth IsLiquid() const { return false; }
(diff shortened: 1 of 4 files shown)


AquariusPower:
I think, activating wizard autoplay for one hour, filtering the log entry to show only spawned golem details, could give a better general idea if it is ok :)

red-kangaroo:
I've since tried it in both autoplay and my games and it seems that golem spawns are now more reasonable both in numbers and power.

PR #562, opened 2019-07-24
Black market

red-kangaroo:
Especially in later stages of the game, players can find themselves with 
large amounts of money that can no longer be spend on anything useful 
(barring priest services and fixing equipment in Attnam), as shops not 
always offer good items, or everything of value was already bought.

This adds a new dungeon, the Black Market, which is randomly placed on 
the worldmap and may be found by players who go explore the wilderness. 
(This also gives some incentive to go explore, as otherwise the player 
will only go for the shortest route between UT, Attnam and GC/TX) It 
has two levels: The first one is full of guards, the second has a huge 
shop of overpriced items.

The purpose of this is to have a place where players can spend all their 
hard-earned money, with enough items generated to ensure there will be 
at least something for every player. However, to prevent unbalancing the 
game with new shopping options, the Black Market has intentionally 
overpriced items that will only be affordable to players who have 
accumulated some excess funds, therefore allowing an end-game
shopping spree.

The shopkeeper and her shop guards were balanced about being next
to impossible to beat even for an end-game player, as the shop holds
too much wealth to give to the player.

Thoughts and polemic are more then welcome.


Main/Source/rooms.cpp
```diff
@@ -700,7 +700,7 @@ truth bananadroparea::PickupItem(character* Hungry, item* Item, int)
 
   if(Hungry->IsPlayer())
   {
-    if(!Item->IsBanana() && !Item->IsLanternOnWall())
```
fejoa: Did you drop the holy banana there by mistake?

Script/dungeons/GloomyCaves.dat
```diff
@@ -269,7 +269,7 @@ Dungeon ELPURI_CAVE;
           e == Random { MinPrice = 50; MaxPrice = 10000; Category = FOOD|POTION; }
           u == Random { MinPrice = 200; MaxPrice = 10000; Category = WAND|TOOL; }
           r == Random { MinPrice = 200; MaxPrice = 10000; Category = SCROLL|BOOK; }
-          s == wand(WAND_OF_STRIKING);
```
fejoa: That's a big change from what players are used to. Maybe leave them both on the same square. Players might not opt to go to the black market and they will still require ascension gear like this wand.

fejoa:
> BTW, why was Tomb of Xinroch "locationAW"?

`locationAW` is part of the generic locations engine. You can re-name it, it won't break anything :)

Main/Source/rooms.cpp
red-kangaroo: No no, bananadroparea is used for Decos' house, so the old code meant that players could take anything inside except for bananas and lanterns without repecussions. Thus they could steal without angering NA.

I changed it to only allowing free taking of quest items (the player would have to bring them there and mistakenly drop them), so players can no longer steal from Decos.

red-kangaroo:
BTW, I set the price of the black market map at roughly 100gp, with the idea that people could either buy something immediately from the dungeon shop, or spend the money on getting to know the black market and buy more stuff later. I don't think the price should be lower that this, but maybe it should be even higher, if we wanted to restrict the access to only the rich? Something like 500gp maybe?

Opinions?

fejoa:
> I set the price of the black market map at roughly 100gp

Sounds good. Is this ready to merge?

red-kangaroo:
@fejoa Yep. :)

PR #563, opened 2019-08-05
Make NPCs talk on their own

red-kangaroo:
Add a small chance that both friendly and hostile NPCs will chat with you on their own, so that the player has a chance of hearing even hostile dialogue from monsters (as nobody would normally chat while under attack). Also makes the towns a bit more lively, as it seems like you are overhearing conversations. ;)

Special cased BeTalkedTo() functions had to be changed to take into account that player no longer needs to be always adjacent for this function to be called.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,6 +62,7 @@ addons:
       - libsdl2-mixer-dev
 
 before_install:
+  - export HOMEBREW_NO_AUTO_UPDATE=1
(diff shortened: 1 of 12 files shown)


fejoa:
Looks pretty good. Any more changes, or is it good to be merged?

red-kangaroo:
@fejoa It should be good to go. :)

PR #587, opened 2020-03-24
Lights (feels ready), new sfx (ready), crafting skill progress (ready)

AquariusPower:
Weaker lights from small rocks feels ok now, but it would be better more extensive testing (ex.: crafting daggers from crystal rocks) before changing the related default (disabled) option.

New sfx for cutting webs, is it ok?

BUGFIX: Also, fixed it when crafting a key that breaks (or anything else that has no BROKEN configuration on .dat files).

And command line --help with some info and functionalities:
--defgen Generate defines validator source file. (NEW)
--defval Validate defines. (NEW)
--version Show current game version. 

**Feel free to make any adjustments, even if minor.**

Obs.: developer console is still broken

PR #595, opened 2020-04-25
Craft Skill

AquariusPower:
It has also a fix for crafting something that breaks and has no config at .dat files.


AquariusPower:
this should receive an update I didnt mange yet to bring from https://github.com/Attnam/ivan/pull/587
about a chance of gaining craft skill every craft turn on long lasting crafting tasks.

red-kangaroo:
@AquariusPower One thing I noticed in #596 : When splitting rocks with no appropriate tool, the game tells you that you need a cutting tool, but you actually need a hammer.

Main/Source/cmdcraft.cpp
```diff
@@ -1592,6 +1731,25 @@ struct srpWorkBench : public srpOltBASE{
...
 };srpWorkBench rpWorkBench;
+struct srpTWorkBench : public srpOltBASE{
+  virtual bool spawnCfg(recipedata& rpd){
+    rpd.otSpawnType=CTT_FURNITURE;
+    rpd.otSpawnCfg=TAILORING_BENCH;
+    return true;
+  }
+
+  virtual void fillInfo(){
+    init("build","a tailoring workbench");
```
red-kangaroo, line 1742: Maybe change normal workbench to "crafting workbench" then, to make it clear in the F menu how are the two workbenches different?

red-kangaroo:
Could we get reorganized skill overview? Right now, it says "Your experience in weapon categories"/"battle bonus", and then lists crafting. :)

There's crafting skill in the @ menu, and then crafting proficiency in F menu. Does crafting skill increase your crafting proficiency?

When crafting from ingots, they are instantly transformed to lumps if you cancel the crafting after selecting materials but before beginning the crafting action. They should only get lumped if you start crafting and then fail/stop.

What if you could craft any furniture, even if they don't do anything? Maybe the player wants to build a small house with some beds and an oven.

Spider silk from webs is a great touch. :)

AquariusPower:
> @AquariusPower One thing I noticed in #596 : When splitting rocks with no appropriate tool, the game tells you that you need a cutting tool, but you actually need a hammer.

if I am not wrong, I initially thought and coded it like: to split a rock you would need a dagger and a hammer, so that you would aim the dagger on the rock and hit the hammer on the dagger's hilt. But the warn/user/help message may be imprecise!

Main/Source/cmdcraft.cpp
AquariusPower: may be "smithing bench" but I read it is more related to metal work, "crafting workbench" is more generic indeed and may be clearer

AquariusPower:
> Could we get reorganized skill overview? Right now, it says "Your experience in weapon categories"/"battle bonus", and then lists crafting. :)

The skill progress system was already implemented as weapon skills, it just took me some more time than usual to realize that xD

> There's crafting skill in the @ menu, and then crafting proficiency in F menu. Does crafting skill increase your crafting proficiency?

in the F menu (that is a good and innocent menu, just meaning craFting xD), so it works like this:
You have the craft skill, that increases with each related action.
You have a crafting bonus skill based on your stats (mainly dexterity).
On that special menu, the crafting value shown is: CraftSkill+CraftBonus

> When crafting from ingots, they are instantly transformed to lumps if you cancel the crafting after selecting materials but before beginning the crafting action. They should only get lumped if you start crafting and then fail/stop.

Agreed! and the way it is now it is quite annoying/unfair...
But to let it work properly, it would require a quite more complex code (I guess), 
or... I just thought it... we could implement an "**UNDO**" code, that would join the lump back into the ingot if the crafting action doesnt really begin. 
Cutting the ingot (or any other crafting basic item) and creating the lump is a preparation to the crafting process but it should be undone if that isnt started!

Cool, I think an UndoCraftingPreparation() could be a good workaround!
May be, the new cut ingot and the new lump could be just deleted and the old one could be just restored, I dont know yet the best way to implement it tho.

> What if you could craft any furniture, even if they don't do anything? Maybe the player wants to build a small house with some beds and an oven.

Nice! it would open a menu listing them all, may be the useful ones on  the top.
The problem is the volume of materials required to build them. May be all "non useful" could have a default small value, may be 2000cm3, 
or we could add a property on the .dat file related to furniture, something like: ToCraftVolumeRequired = 2000, that being default, but later we could specify each, and it would be better.
 
> Spider silk from webs is a great touch. :)

Used in tailoring as sewing material, but we can craft gloves etc from it too, yep it is fun.
After that I avoid killing spiders to let them create webs I can collect. Btw, in my house irl, I get all these inoffensive spiders and put them on the plants to help fighting insects, inoffensive spiders are very cool! xD

Main/Source/cmdcraft.cpp
red-kangaroo: Yeah, "smithing" is metal work and we already have an anvil for that. :)

fejoa:
It compiles and runs on Windows. I haven't tested anything but it seems to work OK. 
Can we merge this yet?

fejoa:
I would like to merge this next. I think it is ok?

PR #599, opened 2020-05-06
String question history

AquariusPower:
By pressing key up/down.
accepts also DELETE key to edit the string.
copy from https://github.com/Attnam/ivan/pull/587 (merged with `meld` on linux)


fejoa:
I think `save.cpp` was edited in the intervening time. @AquariusPower are you able to resolve? I can do it, but I will have to squint my eyeballs to make sure the control flow is correct in those macros 🧐

fejoa:
Upon closer inspection, the code segment on `master` was the latest. I think it will behave as before. I pray for Arch Linux users everywhere.

fejoa:
Works on windows

PR #601, opened 2020-05-08
Developer Tools (fixed, improved, expanded)

AquariusPower:
Developer Tools:
- Moved wizard auto play code to it's own files, and improved it a bit.
- Created cursed developer game mode, where player won't die (neither score), to skip/avoid any conditions governed by wizard mode.
- Expanded commands list on developer console and fixed it (thx @fejoa!).

copy from https://github.com/Attnam/ivan/pull/587, merged with `meld` on linux
I will check if there are other things that can be split from that PR w/o getting overly complicated.
Some splits required new code dependencies that are now in 3+ branches.


fejoa:
@AquariusPower Are you able to update this PR? I get a `error: 'GetUserDataDir' is not a member of 'game'` in `char.cpp` when I try to compile.

Main/Source/char.cpp
```diff
@@ -2533,7 +2513,14 @@ truth character::TestForPickup(item* ToBeTested) const
 
 void character::AddScoreEntry(cfestring& Description, double Multiplier, truth AddEndLevel) const
 {
-  if(!game::WizardModeIsReallyActive())
+  if(game::WizardModeIsReallyActive())
+    return;
+  if(curseddeveloper::IsCursedDeveloper())
+    return;
+  
+  highscore HScore(game::GetUserDataDir() + HIGH_SCORE_FILENAME);
```
AquariusPower, line 2521: I saw the error here:
https://github.com/Attnam/ivan/runs/4004758260?check_suite_focus=true
I think just removing `game::` will solve it like:
`highscore HScore(GetUserDataDir() + HIGH_SCORE_FILENAME);`
but I think there is about 10 other places to do the same fix or it wont fully compile

AquariusPower: I just created this backup: https://github.com/AquariusPower/ivan/tree/DevTools.TempBackup

AquariusPower:
[dungeon.cpp ](https://github.com/Attnam/ivan/runs/4912348130?check_suite_focus=true)
```
/home/runner/work/ivan/ivan/Main/Source/dungeon.cpp:130:57: error: ‘GetAutoPlayMode’ is not a member of ‘game’
  130 |                              Displacement, WHITE, game::GetAutoPlayMode()<2,
```

Some things from wizautoplay.h(.cpp) were merged into Attnam:master in the past, like GetAutoPlayMode() that was copied into game.h(.cpp).
Now, when merging Attnam:master into DevTools, there are some inconsistencies like dungeon.cpp expects GetAutoPlayMode() to be at game.h(.cpp). But DevTools still have these only at wizautoplay.h(.cpp).

I made this backup https://github.com/AquariusPower/ivan/tree/DevTools.TempBackup of this branch.
I will see what I can do.

AquariusPower:
I still havent found a way to install libsdl2-dev on ubuntu20.04 (there is some dep problems that makes no sense to me and would mess everything, it just wont install anyway).
If someone knows how to setup SDL2 development environment on ubuntu20.04 I would like to know :)
So, this means I fixed online only and could not test on my machine. It will compile but I could not test the way I like: run it to see if it is working.

PR #608, opened 2020-06-08
Auto store items on container based on inscribed regex

AquariusPower:
just inscribe a regex in the container
works for pickup and auto-pickup
(split from https://github.com/Attnam/ivan/pull/587 using `meld` folder compare)


red-kangaroo (changes requested):
This is great!

- The message for storing an item in a container should come after the message for picking it up:

Screenshot from 2020-09-28 17-38-23

- Items will be stored in containers even when they are locked. There should be a check if you can open the container before trying to auto-store an item inside.

AquariusPower:
>     * The message for storing an item in a container should come after the message for picking it up:

agreed, if to move the code (that stores the item) becomes too complicated, just moving the message (store it in a var to display later) would suffice.

>     * Items will be stored in containers even when they are locked. There should be a check if you can open the container before trying to auto-store an item inside.

could check if it is locked, and then store only if we have a key to open it. 
To complement that, we should let NPCs get items from unlocked containers making keeping it locked meaningful.
In the other hand, we could grab a locked chest and using this glitch/bug, store things on it w/o having a key.

I think both could become issues!

red-kangaroo:
> To complement that, we should let NPCs get items from unlocked containers making keeping it locked meaningful.

That would be nice.

> In the other hand, we could grab a locked chest and using this glitch/bug, store things on it w/o having a key.

I would consider that bug abuse. :D Also the player might end up with items stored in chests he doesn't have a key for.

fejoa:
Are we good to merge this @red-kangaroo ?

red-kangaroo:
> Are we good to merge this @red-kangaroo ?

I think there should at least be the check not to store items in locked containers. That could bite the player in the proverbial ass quite easily. :)

Main/Source/game.cpp
```diff
@@ -1324,40 +1324,70 @@ int game::RotateMapNotes()
...
+  for(int i=0;i<vit.size();i++){
+    DBGITEM(vit[i],"checkingIfIsAContainer");
+    itemcontainer* itc = dynamic_cast<itemcontainer*>(vit[i]);
+    if(itc){
+      long lRemainingVol = itc->GetStorageVolume() - itc->GetContained()->GetVolume();
+      DBG3(lRemainingVol,itToStore->GetVolume(),itc->GetLabel().CStr());
+      if(lRemainingVol<itToStore->GetVolume())
+        continue;
+      
+      if(itc->IsAutoStoreMatch(itToStore->GetName(DEFINITE))){
```
AquariusPower, line 1366: I think changing this line to
`if(!itc->IsLocked() && itc->IsAutoStoreMatch(itToStore->GetName(DEFINITE))){`
suffices to solve https://github.com/Attnam/ivan/pull/608#issuecomment-1000075184 request

PR #631, opened 2021-09-13
Worldmap experiments

fejoa:


Script/owterra.dat
```diff
@@ -150,7 +150,7 @@ pyramid
   AttachedDungeon = PYRAMID;
   CanBeGenerated = true;
   HideLocationInitially = true;
-  NativeGTerrainTypes = { 1, DESERT; }
+  NativeGTerrainTypes = { 2, DESERT, JUNGLE; }
```
fejoa, line 153: @red-kangaroo would it be acceptable if the pyramid has the flexibility to generate in either desert or a jungle tile? The generator would try to place the pyramid on a desert tiles first, as a matter of preference. On a minority of tiny sized worlds, the sampler might not pick up any desert tiles, so this change gives the generator the option to fall back to jungle terrain. This means the generator does not need to generate as many worlds. Like, ~10 worlds instead of ~100-1000 :)

red-kangaroo: Sure, no problem. :)

red-kangaroo:
This is great! :) Compiles OK on Linux, haven't run into any bugs yet.

When using Pangea, the generator puts all locations on a single "part" of the landmass - eg. if the continent goes north and warps to the south, all locations will only be in the northern part and the sounthern part will be empty, even though it's easily reachable. I understand it's probably a limitation of the current location reachability check?

It seems that the disappearing main manu from #587 was merged into pull request, too. I don't think that feature should be merged in its current state - there are some white artifact pixels when parts of the graphics disappear, plus I would like if it was toggleable in config (that's just me, I don't want my main menu disappearing ;) ).

fejoa:
I think the changing menu background must've been added in #600. 
Probably need to restore menu.png, and put a condition in `void igraph::LoadMenu()` which is in `Main/Source/igraph.cpp`.

Regarding Pangea, the generator uses the underwater tunnel exit as the starting point and fills up locations nearest to that. So that's one complication.
Another one is `void worldmap::CalculateContinents()`. I haven't touched the continent calculator, so it's still the old one. I think it is based on a simple floodfill algorithm, or something like that. The floodfill algorithm would need to be updated so that it checks the other side of the worldmap as well - depending of course on the world's shape. But what is the right stopping criterion? I guess you wouldn't want the floodfill to wrap around the world more than once... (in the cylindrical case - I don't know about a torus)? I think asking the internet how to do a "floodfill on a torus", or a "floodfill on a cylinder", or even a sphere could yield some relevant info.

PR #632, opened 2021-10-07
Fix main menu extra graphics/sounds

red-kangaroo:
Add an option to use (or not) the changing graphics and sounds in the main menu, which was likely prematurely added in #600 and also appears in #587 and #631.

I also removed the extra graphics for now, leaving Menu1 through Menu5 as the same (old) image. The new images had some artifact pixels and weird resizings, so a better version should be used in a release. The images can be found in .devsPrefs/red-kangaroo/TodoGraphics, if anyone needs them.

The sounds work, though. :)


fejoa:
Oops, that was rather forceful of me to push directly to your branch, apologies @red-kangaroo 

It's desirable that the static menu is the default, good work here. I couldn't load the other files when I switched to the changing menu graphics, so I added them here. Works pretty good. I patched the `Menu2.png` file so it has the correct size now.

fejoa (approved):

PR #656, opened 2024-12-11
cmdcraft: fix tailoring crash

bkoropoff:
Mark sewing thread as secondary material.

This fixes an abort [here](https://github.com/Attnam/ivan/blob/master/Main/Source/cmdcraft.cpp#L1185) that I was running into whenever trying to tailor anything.  I'm not sure if this is the correct way to fix it, but it seems to work.


ryfactor (approved):

ryfactor:
Thanks for submitting this bugfix @bkoropoff, and welcome :)

PR #657, opened 2024-12-12
char: fix null deref when trying to remove trap

bkoropoff:
Cutting down a web can try to remove the trap from a character on the square that isn't actually trapped.  Do nothing in this case.


ryfactor:
Thanks. This bug is easily more than 20 years old, and most players are likely to have come across it at least once before.

ryfactor (approved):

PR #658, opened 2024-12-15
craft: fix crash in case recipe uses 3 ingredients

bkoropoff:
My previous PR didn't quite fix the abort, since some recipes (e.g. whips) use a main, secondary, and sewing thread ingredient.  Just use an array with well-defined indices to allow for future expansion.


bkoropoff:
cc @ryfactor

ryfactor (approved):

PR #681, opened 2026-06-16
improved implementation of outlines

zenorogue:
I have noticed that IVAN includes an option to use my boundary mod.

However, it was done by having copies of files (such as `Graphics/Item-outlined.png`) and using them at startup instead of the correct ones. These files were very old and do not include some new stuff added since then (so presumably some graphics would be missing).

This change makes IVAN instead automatically edit the affected files to add the boundaries. I have also changed it so that no restart is required when switching this option.

Note: to work correctly, I had to hardcode the coordinates of large entities; so `igraph.cpp` will still need to be changed if any new large or otherwise non-conforming graphics are added. I tried to take them automatically from the data instead, but it does not seem possible due to how IVAN is structured -- some coordinates are hardcoded in various cpp files.


ryfactor (approved):
This is a welcome fix to a longstanding issue, thank you @zenorogue

PR #684, opened 2026-07-11
Input part 3, and some related changes

zenorogue:
This is yet another bunch of commits mostly focused on improving the user interface.

I will first mention some bugs I have attempted to fix on the way:

* @AquariusPower has started adding mouse support and help (F1) in some situations (such as lists); however, I do not really understand the intention here. I would just make pressing F1 simply send the F1 key to whoever requested the key. But it treated F1 as some "special" key that is sent via `IsRequestedEvent` and the key question would usually have some timeout after which the "default" key would be sent, and requested special keys would be checked. A similar system was used for mouse presses. This is difficult to work with and I see no benefit. I have made that Get_Key() function simply returns `KEY_SPECIAL` or `KEY_MOUSE_EVENT`, these false keys could simply be ignored, or analyzed using `globalwindowhandler::GetLastMouseEvent()` or `IsRequestedEvent()`.
* The aforementioned key timeout function did not work for me correctly -- because it was based on `clock()` which returns the CPU time spent by the current process, and the process does not use up much time when waiting for the next key. Replaced some uses of `clock()` with the SDL function `SDL_GetTicks()`. (I have also noticed that `clock()` is sometimes used as a kind of random number generator, which I did not replace; and in one place the modern C++ `std::chrono` is used.)
* The map command ('m') was very confusing to me -- it apparently did nothing when the option `DungeonGfxScale` ("graphics scaling factor") was set to 1, other than sometimes displaying a strange message "Cannot display a map that is as big as the world!". After I changed the `DungeonGfxScale` setting to 1, I found that it actually displayed a nice map. I see no reason why this should not work with `DungeonGfxScale`=1, so I disabled these checks. (However, for some reasons I do not understand, I had to add extra screen refreshes so that the map disappears from the screen after 'm'ap is closed.)
* Another bug with the map command: "map notes" use the same field as "engravings", they are just prepended by the '#' symbol. It should not be done like this: a "map note" and "engraving" is, diegetically, a different thing, in particular it would be possible to have both in one place. I have left it as is, but when generating 'l'ook descriptions, the map notes were treated as engravings (and displayed together with their '#' symbol) -- I have fixed this.
* right-clicking now cancels list questions. Since mouse in lists works correctly now IMO, I have changed the default setting controlling its feature to "on". There was a bug in function `UseExtraMenuGraphicsChanger` which caused changing this option to also enable/disable mouse -- not sure what @red-kangaroo (#632) meant here, so I simply commented it out.

Now, for the actual new features:
* Made the mouse work in some places. It works in the main menu. It works in the main input (clicking PC waits, clicking an adjacent tile moves, clicking a further tile routes, clicking an equipment slot changes equipment, clicking the silhouette calls ShowKeyLayout). It works in position questions. It works in direction questions.
* Pressing F1 in the also calls ShowKeyLayout. Since it shows help in general, it should show help there too.
* Pressing > or < in position questions now moves to the stairs.
* The '(g)o' command has been extended. You can press '>' or '<' to route to the stairs. You can press 'x' to autoexplore . The "route" command went through ShowMapWorks for some reason which is buggy, I changed it back to use normal PositionQuestion. Also F1 displays help (there are too many commands to fit in the question line).


zenorogue:
Checks fail because `SDL_MouseWheelEvent` has no `mouseX` and `mouseY` fields. It seems the checker uses an old version of SDL2 for some reason (according to https://wiki.libsdl.org/SDL2/SDL_MouseWheelEvent these fields were added in 2.26.0).

ryfactor:
Ok I have made a note to keep SDL2 up to date in the GH Actions (probably will try to update the runners to Ubuntu-latest if it will cooperate).
Will add a check in the CMakefiles to ensure SDL2 version is > 2.26.0.

ryfactor (approved):

PR #685, opened 2026-07-11
Portability

zenorogue:
Various commits intended to improve portability, in various ways.

Note: I am cross-compiling the Windows version on Linux using `x86_64-w64-mingw32-g++`. So this is 64-bit. I am not sure how to make `cmake` work (I am compiling using my own scripts instead which I have not included in the PR).

* The saves in the Linux and Windows versions I compiled were not compatible, because on Linux "long" is 64-bit, while in Windows, at least according to the x86_64-w64-mingw32-g++ compiler, it is 32-bit, so Windows would read 32 bit of a Linux long and the next 32 bits would be assumed to refer to something else. I have fixed this by making save.h use 64 bits anyway when recording "long" or "ulong". An alternative solution would be redefine "ulong" to be 64-bit and replace all occurrences of "long" with a new 64-bit type.
* The Linux version was configured to install itself in the system -- added a cmake option PORTABLE_BUILD to create a "portable build" working from a single directory. Also this option uniformizes the highscore filename between Linux and Windows. It was mostly for Steam but I guess it could also be used if you wanted to launch IVAN from a USB stick that you use on both Linux and Windows, and keep user data between them.
* There was a problem with `festring` and `fearray` allocating an `int*` before the actual content and then treating it as `ulong`, which did not work when `int*` and `ulong` have different sizes. (Weird that this worked on Wine for me. It crashed on actual Windows though.)
* Also it appears that `_findfist` crashes in w64-mingw32 (even in my minimal program) -- I have added an option `USE_OPENDIR` to use the Unix standard functions `opendir`/`readdir` which work there.

### DIFF
--- a/xbrzscale/libxbrzscale.h
+++ b/xbrzscale/libxbrzscale.h
@@ -16,6 +16,9 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#ifndef _XRBZSCALE_
+#define _XRBZSCALE_
+
 #ifndef XBRZLIB_RELATIVEPATHSDL
 #include <SDL_stdinc.h>
 #else
@@ -49,3 +52,4 @@ class libxbrzscale
   static bool bFreeInputSurfaceAfterScale;
   static bool bFreeOutputSurfaceAfterScale;
 };
+#endif
--- a/xbrzscale/libxbrzscale.h
+++ b/xbrzscale/libxbrzscale.h
@@ -17,7 +17,7 @@
  */
 
 #ifndef _XRBZSCALE_
-#define _XRBZSCALE_
+#define _XBRZSCALE_
 
 #ifndef XBRZLIB_RELATIVEPATHSDL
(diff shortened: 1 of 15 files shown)


xbrzscale/libxbrzscale.h
```diff
@@ -16,6 +16,9 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#ifndef _XRBZSCALE_
```
ryfactor:
```suggestion
#ifndef _XBRZSCALE_
```

xbrzscale/libxbrzscale.h
```diff
@@ -16,6 +16,9 @@
...
+#define _XRBZSCALE_
```
ryfactor:
```suggestion
#define _XBRZSCALE_
```

ryfactor (changes requested):

ryfactor (approved):

